PR #16, opened 2016-06-09
more final viz.yaml structure

aappling-usgs:
@jread-usgs here's what we've decided on, at least for the near future


inst/testviz/viz.yaml
```diff
@@ -57,56 +60,52 @@ sections:
...
-        mimeType: text/tab-separated-values
-    buildfile: scripts/3_visualize/landUse.R
-    type: R
-    tags: [mobile, desktop, ie]
+    location: relativeAbundanceFig.svg
+    depends: mungedRelativeAbundance
+    visualizer: visualizeRelativeAbundance
+    mimeType: image/svg+xml
+    export: false
+images:
```
jread-usgs: are images anything that would be in the images dir, including background images? Assuming so, but wanted to check.

aappling-usgs: that's my understanding, yes

PR #17, opened 2016-06-13
makeMakefile('fetch')

aappling-usgs:
I think makeMakefile('fetch') works - at least, it seems OK on my computer.

To see/run the example: Check out `inst/testviz/viz.yaml'. Then run this in R:
```r
vizlab::makeMakefile('fetch')
```

This creates `vizlab/make/fetch.make` and other directories that will be needed by the makefile, including `cache`, `vizlab/make/log`, and `vizlab/make/timestamps`. Then open a bash shell (e.g., alt-t-s) and run this:
```
cd inst/testviz
make -f vizlab/make/fetch.make
```
and watch the makefile first run everything, then (on rerun) only run the check for the sciencebase timestamp (Cuyahoga_timestamp) from there on out.

 I also started working on makeMakefile('process'), and plan to eventually work on makeMakefile('visualize'), but I've run out of time. someone else (@jiwalker-usgs ? @jread-usgs ? @lawinslow ?) may have to take over for a while because I'm committed to other projects for the next two weeks.


jiwalker-usgs:
Cool, lots of what we talked about is in here!  I'll push on with the makefile building for the other parts.

PR #18, opened 2016-06-13
make skeleton of new viz

jiwalker-usgs:


R/viz.skeleton.R
```diff
@@ -0,0 +1,54 @@
...
+#'
+#' @param name character vector containing viz name
+#' @param path character vector of path to create skeleton
+#' @examples
+#' path <- file.path(tempdir(), "test")
+#' viz.skeleton(name="test", path=path)
+#' file.remove(path) # cleanup
+#'
+#' @export
+viz.skeleton <- function (name = "cool-viz", path = ".") {
```
aappling-usgs: i think we were headed toward a convention of camelCase for functions, period.separated for variables?

R/viz.skeleton.R
```diff
@@ -0,0 +1,54 @@
...
+  message("Creating directories ...", domain = NA)
+  safe.dir.create(path)
+  safe.dir.create(file.path(path, "data"))
+  safe.dir.create(file.path(path, "images"))
+  safe.dir.create(layout_dir <- file.path(path, "layout"))
+  safe.dir.create(file.path(layout_dir, "css"))
+  safe.dir.create(file.path(layout_dir, "js"))
+  safe.dir.create(file.path(layout_dir, "templates"))
+  safe.dir.create(script_dir <- file.path(path, "scripts"))
+  safe.dir.create(file.path(script_dir, "1_fetch"))
```
aappling-usgs: i'm up in the air again about script directory names: should they have numbers or not? if so, maybe the corresponding blocks in viz.yaml should also have numbers, as should the folders in cache?

R/viz.skeleton.R
jiwalker-usgs: Good call, I was parroting package.skeleton, but changing makes sense.

R/viz.skeleton.R
```diff
@@ -0,0 +1,54 @@
...
+  safe.dir.create(file.path(script_dir, "read"))
+
+  message("Creating viz.yaml ...", domain = NA)
+  viz.yaml <- file(file.path(path, "viz.yaml"))
+  cat("vizlab: ", "\"", as.character(packageVersion(getPackageName())), "\"\n",
+      "info:\n",
+      "  id: ", name, "\n",
+      "  name: ", "Visualization long name","\n",
+      "  date: ", format(Sys.Date(), "%Y-%m-%d"), "\n",
+      "  publish-date: ", format(Sys.Date() + 14, "%Y-%m-%d"), "\n",
```
aappling-usgs: =) at the `+ 14`

R/viz.skeleton.R
```diff
@@ -0,0 +1,54 @@
...
+      "  path: ", "/", name, "\n",
+      "  description: >-\n",
+      "    Describe project here\n",
+      "pages:\n",
+      "js:\n",
+      "css:\n",
+      "sections:\n",
+      "data:\n",
```
aappling-usgs: instead of data and figures, how about images, fetch, process, and visualize, as in inst/testviz/viz.yaml?

jiwalker-usgs: yeah, I'll switch that over, I was looking at an old file.

R/viz.skeleton.R
jiwalker-usgs: I'd be in favor of removing the numbers from the directories over letting them spread too far.

aappling-usgs: sounds good to me. let's do it. no numbers. we'll communicate the order of processing somewhere else, e.g., in help files

PR #20, opened 2016-06-13
Explaining use of suggests and adding badges

jiwalker-usgs:


README.md
```diff
@@ -20,6 +21,8 @@ The most cutting edge version of the package can be installed if you have the `d
 devtools::install_github("USGS-VIZLAB/vizlab")
 ```
 
+Some packages are only suggested and will need to be installed manually to open up that functionality.  One example is sbtools which can be used if project data is stored on sciencebase.
+
```
aappling-usgs: good idea

PR #21, opened 2016-06-16
added fetchData.url

lawinslow:
@jiwalker-usgs @aappling-usgs Want to take a look? Pairs with the climate-fish-hab PR.


aappling-usgs:
looks great to me. i'll wait a sec to give @jiwalker-usgs time to look, but i think this is ready for merging.

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: vizlab
 Type: Package
 Title: Utilities for building online data visualizations
-Version: 0.1.0
+Version: 0.1.1
```
jiwalker-usgs: I wouldn't worry too much about incrementing the version yet, but :+1:

jiwalker-usgs:
:+1: what do you think of this pattern in general?

DESCRIPTION
lawinslow: I like to stay in the habit.

lawinslow:
Looks good to me. Should the PR be failing?

jiwalker-usgs:
Still need to fix the build

lawinslow:
Ah, got it.

R/fetchData.R
```diff
@@ -62,3 +62,21 @@ fetchData.sciencebase <- function(viz.id, data.info) {
...
   invisible()
 }
+
+#' \code{fetchData.url} downloads a file from the specified URL
+#' 
+#' @param viz.id the identifier for this data item in viz.yaml
+#' @param data.info content information for this viz.id from the viz.yaml
+#' @rdname fetchData
+#' @export
+fetchData.url <- function(viz.id, data.info){
```
jread-usgs: `@importFrom httr GET` ?

lawinslow: What's the strategy with this package. Sbtools is just suggests so I was leaving this suggests as well.

jiwalker-usgs: Yeah, I started using suggests to keep the actual dependencies to a minimum.  I'm not sure if that makes things easier or harder.

lawinslow: I say bounce them both to imports. sbtools should be on CRAN soon. I always like it when I can keep a CRAN-only imports list.

Issue #5, opened 2016-06-06
Generated remake.yaml files


jiwalker-usgs:
Much of the remake is boilerplate that can be generated from viz.yaml declarations already made.  Define what can be generated and what should be left to the user.

jread-usgs:
how far does this go? to be decided. At least figs, data, and layout

aappling-usgs:
i will take first stab at this. looks like i don't have permission to adjust Assignees, etc. - @jiwalker-usgs , may i have permission, please?

jiwalker-usgs:
I assigned you, I'll check on permissions.

aappling-usgs:
For discussion (@jiwalker-usgs, @jread-usgs, etc.): 

We're currently building a system where arguments to any function get passed in as named arguments in viz.yaml. For example: 
```
fetch:
  -
    id: Cuyahoga
    location: cache/fetch/CuyahogaTDS.csv
    fetcher: sciencebase
    remoteItemId: 575d839ee4b04f417c2a03fe
    remoteFilename: CuyahogaTDS.csv
    mimeType: text/csv
```
`remoteItemId` and `remoteFilename` are expected by `fetchData.sciencebase` and are accessed as `getContentInfo(viz.id, 'fetch')$remoteItemId`, etc.

This is sort of nice because the arg structure is flat in viz.yaml, and args like `remoteFilename` could also apply to some other functions that might get written in the distant future such as `writeData.sciencebase`. But it gets strange when you start thinking about custom fetchers, processors, and visualizers:
* do these custom functions also get args passed in as named fields in the viz.yaml data item? If so, how do we prevent users from accidentally creating an argument name that overwrites some basic information about the item (e.g., `reader`, `mimeType`, or `export`)? 
* my current implementation passes the entire output of `getContentInfo(viz.id, 'fetch')` to every fetcher implementation by default. This guarantees that any custom arguments will reach the fetcher, but we're also providing more information to custom functions than I think is strictly necessary - for example, _should_ `fetchData.mycustomfetcher` have easy access to `data.info$reader`? It's possible but nontrivial to filter out `reader` and other non-custom fields when all fields are in the same flat list for each item.

Alternatives to having named arguments within the main fields of a viz.yaml data item:

[...]

PR #25, opened 2016-06-20
switch to nested args in viz.yaml

aappling-usgs:
NEW for fetchData, processData, and visualizeData:
(@lindsaycarr, @jread-usgs, @jiwalker-usgs, @lawinslow)

1. All args given in the `args` block of a viz.yaml item are passed through by their names to their fetch/process/visualize function.
2. Args in the `args` block are placed in quotes - i.e., they're all characters, consistent with other fields in the viz.yaml. Convert and/or evaluate them within your functions as needed.
2. fetch/process/visualize functions can no longer expect to be handed a `data.info` arg
3. fetchData.sciencebase and fetchTimestamp.sciencebase and fetchData.url need their args to be within the `args` block rather than on the same level as other info for an item.

Example:
```
# viz.yaml
fetch
  -
    id: Cuyahoga
    location: cache/fetch/CuyahogaTDS.csv
    fetcher: sciencebase
    args:
      remoteItemId: 575d839ee4b04f417c2a03fe
      remoteFilename: CuyahogaTDS.csv
    mimeType: text/csv
```

```
# function definition
fetchData.sciencebase <- function(viz.id, remoteItemId, remoteFilename, ..., outfile) {
  ...
}
```

```
# fetch.make
cache/fetch/CuyahogaTDS.csv: vizlab/make/timestamps/Cuyahoga.txt
	${RBATCH} "--args fun=fetchData funargs=list(viz.id='Cuyahoga',remoteItemId='575d839ee4b04f417c2a03fe',remoteFilename='CuyahogaTDS.csv',outfile='cache/fetch/CuyahogaTDS.csv')" \
	vizlab/make/callFunction.R vizlab/make/log/fetch/Cuyahoga.Rout
```

based on recent use cases and discussion in #5


lawinslow:
args block seems good to me.

lindsaycarr:
Sounds like we are good to merge @aappling-usgs ?

aappling-usgs:
great, let's do it.

Issue #33, opened 2016-07-02
ability to switch on/off the checking of remote timestamps


aappling-usgs:
It's annoying/time consuming to wait for the fetchTimestamps.sciencebase call every time you want to remake a figure. On the other hand, it could sometimes be frustrating if we didn't check the timestamps, because if something changed on sciencebase and you didn't know it, the updates wouldn't reflect that change. So make a switch to decide whether to always check or never check.

PR #34, opened 2016-07-02
fix #32, #33

### LINKED ISSUE #32: fetchData('vizid'), processData('vizid'), visualizeData('vizid')
implement an option to pull the arguments other than `viz.id` from the viz.yaml if the user doesn't specify them directly. this should help with rapid iteration on a specific function such as `visualizeData.myThirdDataset()`

aappling-usgs:
@jread-usgs, @lindsaycarr - do these changes resolve your needs to (1) turn off SB timestamp checking and (2) iterate on fetch, process, and visualize functions faster? I think/hope they will. 

@jiwalker-usgs - you good with the choice to name the timestamp checker flag `fetchTimestamp` in viz.yaml? also note switch from xxData.default to xxData.character - I think this is a more R-esque approach (and saves a few code lines) but couldn't see/remember how to do it until today.

all, and maybe especially @ldecicco-USGS b/c you've had wise cautions and @lawinslow b/c you have those multi-file zip files, wondering about design/presence of `readData.filepath` and `readData.folder` (i just now added .filepath and we added .folder pretty recently). The effect is to free up data-reading to be done outside of the readData framework. The question is whether this is the balance we want between complexity and utility. 

1. It's more complex because now there's more than one way to read data (roll a new readData.xxx versus read the data within a processData or visualizeData function) and the makefile has less precise awareness of the dependencies. 
2. It seems to be useful when a visualize function depends on a bazillion files inside a zipfile or folder and you don't want to specify all the individual file dependencies, as is the case for the walleye-bass climate files, but I'm not yet clear on the full range of use cases. 

Input welcome.


jread-usgs:
@aappling-usgs I wasn't able to figure out how to use the #32 additions in this PR. 

I tried a couple of different ways, but end up with a recursive error for them. From main dir in `climate-fish-habitat`
```r
visualizeData('processedWallyTrends')
autoargs: overriding manual & default args with viz.yaml args
Sourcing scripts:
   scripts/process/processWalleyeDecline.R

Attaching package: ‘dplyr’

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
Error during wrapup: evaluation nested too deeply: infinite recursion / options(expressions=)?
```

aappling-usgs:
rats, thanks for testing. maybe we can debug at your computer for a few minutes tomorrow.

small working example 1 from the `vizlab` project:
```r
> setwd('inst/testviz')
> library(vizlab)
> eg1 <- processData('CuyahogaShort')
autoargs: overriding manual & default args with viz.yaml args
Sourcing scripts:
   scripts/process/cuyahoga.R
> eg1r <- readData('CuyahogaShort')
> head(eg1r)
  MONTH    TIME TDS    Q
1    11 1978.87 710  245
2    12 1978.96 430  736
3     1 1979.04 410  508
4     2 1979.12 700  578
5     3 1979.21 260 4590
6     4 1979.29 260 4670
```

small working example 2 from the `vizlab` project:
```r
> eg2 <- visualizeData('relativeAbundanceFig')
autoargs: overriding manual & default args with viz.yaml args
Sourcing scripts:
   scripts/visualize/relativeAbundance.R
Sourcing scripts:
   scripts/read/myCsvReader.R
> file.info(getContentInfo('relativeAbundanceFig')$location)
                                  size isdir mode               mtime               ctime               atime exe
figures/relativeAbundanceFig.svg 31219 FALSE  666 2016-07-04 20:03:32 2016-06-19 15:45:52 2016-06-25 08:24:20  no
```

aappling-usgs:
@jread-usgs - I tried out your `climate-fish-habitat` example. You were calling `visualizeData` on a `process` item instead of a `visualize` item. I've added error checking to break more quickly and clearly in such cases. Give it another spin?

now if you call visualizeData on a non-visualize item:
```r
> visualizeData('processedWallyTrends')
Error in getContentInfo(viz.id, block = "visualize", no.match = "stop") : 
  no matches for viz.id 'processedWallyTrends' in block[s] visualize
```

and if you didn't define a processor for a process item (or a visualizer for a visualize item, etc.):
```r
> processData('processedBassTrends')
autoargs: overriding manual & default args with viz.yaml args
Sourcing scripts:
   scripts/process/processWalleyeDecline.R
Error in processData.character("processedBassTrends") : 
  please specify a processor for viz.id 'processedBassTrends' in viz.yaml
```

jread-usgs:
Awesome, thanks! I am going to give this a spin in the next couple of hrs.

jread-usgs:
This looks good to me! 👍

jread-usgs:
We waiting on others for merge? I think lawinslow will be out for a while, but others are around

jiwalker-usgs:
I was thinking the about `fetchData.character` et al. for a bit and I do think it makes more sense.  We probably want to bring back a `fetchData.default` for a fall through at the end when there is no handler for a class.  I think now there might be a .none in place there.

jiwalker-usgs:
For turning off the timestamp feature, I like having it on individual items, but we might want an environment variable to turn it off everywhere.  I could see not wanting to run it for rapid development, but still running it in CI.

jiwalker-usgs:
This looks good to me :+1:

lindsaycarr:
Any way to change the percentage coverage can change and still make it pass? I hate seeing all these little red x's on perfectly good PRs

aappling-usgs:
thanks, all.

@jiwalker-usgs :
* i actually like not having a `fetchData.default` because the current fall-through (when there is no default) tells you that no specific fetchData method is available, which I think is appropriate. e.g., if you misspell the fetcher for MayflyNymph:
```r
> fetchData('MayflyNymph')
autoargs: overriding manual & default args with viz.yaml args
Sourcing scripts:
   scripts/fetch/mayflyNymph.R
Error in UseMethod("fetchData") : 
  no applicable method for 'fetchData' applied to an object of class "mayflyNydmph"
```
I guess the above error message could be made even clearer, but before this PR i'd had a few lines of code in every .default function just to give a very similar error. still, i'd be glad to have a `fetchData.default` if there were a fall-back function that made sense for all fetchable data.
* an environment variable seems like a great idea for turning on/off the timestamp checks. will leave the viz.yaml flag in place for now but totally willing to see it replaced later.

@lindsaycarr - yeah, sorry about the coverage. any ideas (anybody) how to CI-test functions like these that rely on a whole suite of files & folders?

PR #37, opened 2016-08-12
Refactor

jiwalker-usgs:
Cleanup might have missed some stuff, and there might be some bits missing, but in general it seems to work.  It is still lacking in unit tests, so if we are good with this pattern I'll start working those in.


R/makeMakefiles.R
```diff
@@ -24,28 +24,28 @@ makeMakefiles <- function(blocks=c('fetch','process','visualize'), display=FALSE
...
+makeTopMakefile <- function(outfile='Makefile') {
   makefile <- paste(
     paste0('# Makefile for overall visualization project\n# Do not edit (autogenerated)'),
     makeMakeMacros(),
     makeMakeRunmakeRules(),
     makeCleanRule(),
     sep='\n\n')
   writeLines(makefile, con=outfile)
-  if(!dir.exists('vizlab/make/log/make')) dir.create('vizlab/make/log/make')
+  if(!dir.exists('vizlab/make/log/make')) dir.create('vizlab/make/log/make', recursive=TRUE)
```
jread-usgs: 👍

R/makeMakefiles.R
```diff
@@ -257,83 +257,92 @@ makeMakeItem.fetch <- function(item.info, ...) {
...
 #' @export
 makeMakeRulePair <- function(item.info, block, concat=TRUE) {
-  
+
   # arg prep
   squote <- function(x) paste0("'", x, "'")
   dquote <- function(x) paste0('"', x, '"')
   data.file <- item.info$location
-  if(grepl(" ", data.file)) data.file <- dquote(data.file)
+  if(!is.null(data.file) && grepl(" ", data.file)) data.file <- dquote(data.file)
```
jread-usgs: what happens if it is null?

R/publish.R
```diff
@@ -0,0 +1,187 @@
...
+  return(viz)
+}
+
+#' Image publishing
+#'
+#' @param viz vizlab object
+#' @rdname publish
+#' @export
+publish.img <- function(viz) {
+  required <- c("alttext", "relpath")
```
jread-usgs: 👍 💯

R/publish.R
```diff
@@ -0,0 +1,187 @@
...
+#'
+#' @param viz vizlab object
+#' @rdname publish
+#' @export
+publish.js <- function(viz) {
+  required <- c("relpath", "mimetype")
+  viz <- NextMethod()
+  checkRequired(viz, required)
+
+  output <- sprintf('<script src="%s" type="text/javascript"></script>', viz[['relpath']])
```
jread-usgs: does it not work to do this as `'<script src="%s" type="text/javascript"/>'`?

R/publish.R
```diff
@@ -0,0 +1,187 @@
...
+#'
+#' @param viz vizlab object
+#' @rdname publish
+#' @export
+publish.css <- function(viz) {
+  required <- c("relpath", "mimetype")
+  viz <- NextMethod()
+  checkRequired(viz, required)
+
+  output <- sprintf('<link href="%s" rel="stylesheet" type="text/css">', viz[['relpath']])
```
jread-usgs: does `<link>` get closed out somewhere else?

R/publish.R
```diff
@@ -0,0 +1,187 @@
...
+#'
+#' @param viz vizlab object
+#' @rdname publish
+#' @export
+publish.svg <- function(viz) {
+  required <- c("relpath")
+  viz <- NextMethod()
+  checkRequired(viz, required)
+
+  output <- sprintf('<object id="%s" type="image/svg+xml" class="svgFig" data="%s"></object>',
```
jread-usgs: are we considering a method for inline svgs? (not as objects)

R/makeMakefiles.R
```diff
@@ -342,8 +351,8 @@ makeMakeRulePair <- function(item.info, block, concat=TRUE) {
   rules$file.data <- makeMakeBatchRule(
     target=data.file,
     depends=dep.files,
-    fun=paste0(block, 'Data'),
-    funargs=c(viz.id=squote(item.info$id), dep.args, sapply(unlist(item.info$args), squote), outfile=squote(item.info$location)),
```
aappling-usgs: how do data file dependencies get handled now?

R/publish.R
jiwalker-usgs: This was copied from existing index file, should be self-closing 

good catch :football:

R/publish.R
jiwalker-usgs: Yeah, I looked at SVGInjector but haven't tested it.  We could either use that or just read and write here.  I just went with this to not change everything.

R/publish.R
jiwalker-usgs: I think that breaks firefox or IE or something

jiwalker-usgs: weird isn't it

jread-usgs: it is

R/makeMakefiles.R
jiwalker-usgs: This is probably not working as is, I'm trying to simplify the Makefile to function call transaction by moving the handling into the vizlab functions, but I might be missing out on some of the benefits of having it in the Makefile.  What do you think the makefile blocks should look like for this?

aappling-usgs: Hmm...maybe I just didn't think about it enough on the first go-round, because now I see that the `depends` argument is unchanged, which is really the bit that worried me before - I had thought you were no longer telling the makefile to check & remake dependencies of plots, etc. But that's false, isn't it - you're still declaring dependencies. 

The part I'm coming to terms with now is that the makefile doesn't automatically pass dependencies to the visualizer/processor. I don't love that the visualize.plotX() function that the user writes is responsible for calling something like `data <- lapply(viz[['depends']], readData)` to pull in the data dependencies, because that leaves somewhat more on the user's shoulders than before...but it's not bad. It's definitely more flexible, for better and worse - e.g., if you want to read the same data file in different ways in different visualizers, now you can just call fread() or whatever directly in your process/visualize function. So I'm OK now.

Issue #39, opened 2016-08-23
Read- Mime types- Support more file types [3]


eread-usgs:
Read. Support more file types, and more support (to a limited extent), and make more extensible so that people can bring their own files. 

Currently mime types are hard coded, and we only support 5.

eread-usgs:
There are a couple places that mime types are relied upon. So centralizing that in some reasonable way is the preference. Right now Fetch, Read, and Publish all use mime types.

eread-usgs:
@jiwalker-usgs suggests take the mime type dictionary and pull it out into a config file so it's editable by a user of vizlab. Right now the dictionary is a switch statement. 

There is a mime type field in the yaml that specifies mime type.

jiwalker-usgs:
The switch statements that map mimetype to reader and mimetype to publisher should be combined and put in an external file (`inst/etc/mimemap`?).  We should then add this to the skeleton process so each viz has a copy of it that can be modified to add or override mappings that occur.  If this file does not exist in that spot, we can fall back to the `inst/etc/mimemap` or whatever it gets called.

eread-usgs:
- [ ] what to do with zip files- will return a file path, processing step needs to unzip
- [ ] new reader type- readdata.dir for directories. use readR for this. use package mime, function guess type
- [ ] making a lookup table for mime types instead of using the switch statement.

jiwalker-usgs:
fixed with #53

PR #43, opened 2016-08-24
mimetype WIP

lindsaycarr:
See #39


R/readData.R
```diff
@@ -22,11 +22,11 @@ readData.character <- function(viz) {
   readData(viz)
 }
 
-#' \code{readData.csv} reads a csv file.
+#' \code{readData.txt} reads a text file.
 #'
 #' @rdname readData
 #' @export
-readData.csv <- function(viz) {
+readData.txt <- function(viz) {
```
lindsaycarr: thought this should be a more generic name since it is using fread and can be tab or some other delimiter

R/readData.R
```diff
@@ -93,11 +93,10 @@ as.reader <- function(viz, ...) {
     mimetype <- viz[['mimetype']]
     reader <- switch(
       mimetype,
-      "text/csv" = "csv",
-      "text/tab-separated-values" = "csv",
+      "text/csv" = "txt",
+      "text/tab-separated-values" = "txt",
       "text/yaml" = "yaml",
       "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" = "excel",
-      "application/zip" = "filepath",
```
lindsaycarr: seems strange to support "application/zip" if all we do is throw the filepath back. I think a warning should pop up if you have a zip file that says you need to unzip it OR we figure out how to unzip and return the new folder ourselves.

R/readData.R
jiwalker-usgs: maybe we should go with something more descriptive (delim?) in case we want to handle non-delimited text/plain?

R/readData.R
jiwalker-usgs: I think there is a `process.zip()` that will unzip the file, so this is meant to say that we don't know how to read a zip as it is somewhat arbitrary.

R/readData.R
lindsaycarr: readData.tabular ? because readData.txt could be a straight text file where the output is an R string vector

jiwalker-usgs: yeah, I think tabular makes sense

R/readData.R
```diff
@@ -74,6 +74,14 @@ readData.folder <- function(viz){
   dir(viz[['location']], full.names=TRUE)
 }
 
+#' \code{readData.txt} returns names of files inside a folder
+#'
+#' @rdname readData
+#' @export
+readData.txt <- function(viz){
+  scan(viz[['location']], what="character", sep="\n")
```
lindsaycarr: this makes a lot of assumptions

NAMESPACE
```diff
@@ -35,11 +35,11 @@ S3method(publish,resource)
 S3method(publish,section)
 S3method(publish,svg)
 S3method(readData,character)
-S3method(readData,csv)
 S3method(readData,excel)
 S3method(readData,filepath)
 S3method(readData,folder)
 S3method(readData,rds)
+S3method(readData,txt)
```
jiwalker-usgs: need to update NAMESPACE with tabular

lindsaycarr:
do we care about XML? Probably not.... @jiwalker-usgs ?

R/readData.R
jiwalker-usgs: cool, I usually just use readLines, do you know the pros/cons between them.  Also, thinking about this again, we can probably use this to read in mustache template files more clearly (I was doing direct readLines calls)

R/readData.R
```diff
@@ -93,11 +101,11 @@ as.reader <- function(viz, ...) {
     mimetype <- viz[['mimetype']]
     reader <- switch(
       mimetype,
-      "text/csv" = "csv",
-      "text/tab-separated-values" = "csv",
+      "text/csv" = "tabular",
+      "text/tab-separated-values" = "tabular",
       "text/yaml" = "yaml",
+      "text/paragraph" = "txt",
```
jiwalker-usgs: This will probably be a few different mimetypes, I'd say text/plain should be the main one though as mentioned above it could also handle text/template or text/mustache.

jiwalker-usgs:
I think xml would be good to handle, but it is probably a bit complex for a good general solution at this time, so we can handle it as we get an example that makes sense.

R/readData.R
lindsaycarr: by "it" do you mean this pkg?

jiwalker-usgs: yes, the `readData.txt()` could handle all those mimetypes

Issue #48, opened 2016-09-08
note that make should be used only in the context of makefile, not as a verb for function name


eread-usgs:

wdwatkins:
done in #49

PR #49, opened 2016-09-09
Changed function names, other bug fixes

wdwatkins:
- Fixed a few random things that came up in the course of testing
- Switched verbs in function names from `make` to `create` #48 
- fetch url and file timestamp functions are there, but not being utilized yet


R/fetchTimestamp.R
```diff
@@ -63,9 +63,32 @@ fetchTimestamp.sciencebase <- function(viz) {
...
+#' 
+#'  @rdname fetchTimestamp
+#'  @importFrom httr HEAD
+#'  @importFrom httr headers
+#'  @export
+fetchTimestamp.url <- function(viz) {
+  
+  #How will a url be recognized?  Specified in yaml
+  url <- viz$location
+  tag <- headers(HEAD(url))$`last-modified`
```
jiwalker-usgs: I've been trying to get in the habit of using [[]] rather than $ (I think JR and hadley convinced me of that), and this is one case where it is probably more clear.

R/fetchTimestamp.R
```diff
@@ -63,9 +63,32 @@ fetchTimestamp.sciencebase <- function(viz) {
...
+  fileLoc <- viz$location
+  #get date file timestamp
+  time <- file.info(fileLoc)$mtime
+  return(time)
 }
 
+#' check a URL for timestamp
+#' 
+#'  @rdname fetchTimestamp
+#'  @importFrom httr HEAD
```
jiwalker-usgs: I think httr might need to be added to DESCRIPTION

jiwalker-usgs: Moved from suggests to imports rather.

R/fetchTimestamp.R
aappling-usgs: what's your reasoning for [[]] instead of $?

jiwalker-usgs: I don't remember either JR or hadley's reasoning (it was somewhere in hadley's talk at useR about practical vs. utopia).  But in this case I think the $` might be cleared up by switching.

wdwatkins: Looks like `[[]]` makes it cleaner if you wanted to use a variable containing the attribute name as well

jiwalker-usgs: I just remembered it, it has to do with partial matching
```
foo <- list(bar=42)
foo$a # NULL
foo$b # 42
foo[['b']] # NULL
```

Issue #40, opened 2016-08-23
Fetch- timestamp [2]


eread-usgs:
Part of fetch. Timestamp function is working for SB, but there is more to be done for timestamping done on local files, URLs, or other methods.

Rather than downloading the file each time, timestamp is inspected and service is checked to determine if there's a change. timestamp.sciencebase function does this for SB. 

Needs to be tested and verified. There are three Fetch functions implemented, for each of those there's a timestamp function that needs to be checked. 

Fetch: File, SB, and URL functions. 

URL needs to be implemented, SB needs to be tested and verified. File needs to be tested.

eread-usgs:
timestamp for file is taken care of. URL timestamp is more complicated. 
code to retrieve timestamp is in place, but nothing is using it. 

- [ ] override for timestamp
- [ ] The logic needs to be put in place to check the timestamp. 
- [ ]  URL timestamp

wdwatkins:
The functions themselves should be good to go in #52, but some other issues:

- For local files, there is some redundancy in checking the timestamp, since `make` already checks the modified tag itself when the data file is listed as a dependency.  Do we want to only use the timestamp as a dependency to avoid this?

- What happens when `fetchTimestamp` is false in the viz.yaml?  It skips the fetch, but what if as in the example viz iris-data there is no processed file cached?  If you wanted to change the visualization step, the fetch would still have to happen.  Should there be 'skip' flags in each section, so that they can be turned off if you want?  

- An example with a local file that gets processed would make the things☝️  easier to test

eread-usgs:
More examples in the Process script would help to show this works. Add another figure to the test viz that accesses something remote. Cuyahoga example.

wdwatkins:
#61 should finish this.  Also made a PR to the example viz repo to add a sciencebase block https://github.com/USGS-VIZLAB/example/pull/7

eread-usgs:
fixed in #61

PR #52, opened 2016-09-15
completed fetchTimestamp functions

wdwatkins:
The fetchTimestamp functions are functional, and have tests.  There are still unresolved questions though with how these functions will be utilized though.  See #40 

Some other bugs that came up in the course of running the `example-viz` are also fixed here


R/fetch.R
```diff
@@ -20,9 +20,13 @@ fetch.character <- function(viz) {
   # all.args <- getAutoargs(data.info, fun='write')
 
   viz <- as.fetcher(viz)
-
+  #check timestamp
+  #if(viz$fetchTimestamp){
```
jiwalker-usgs: We should avoid checking in comments.

R/fetchTimestamp.R
```diff
@@ -54,21 +55,31 @@ fetchTimestamp.sciencebase <- function(viz) {
   # write the new timestamp to the file
   if(!is.na(new.timestamp) && (is.na(old.timestamp) || (new.timestamp != old.timestamp))) {
     writeTimestamp(new.timestamp, locateTimestampFile(viz[['id']]))
+    return(TRUE)
```
jiwalker-usgs: Modify documentation to describe the return from this function.

R/fetchTimestamp.R
```diff
@@ -108,14 +137,18 @@ writeTimestamp <- function(new.timestamp, outfile) {
 #' @return character vector location of timestamp file
 locateTimestampFile <- function(id) {
   # TODO standardize timestamp file location
+  #vizlab/make/timestamps?
+  timestampDir <- "./vizlab/make/timestamps"
+  timestampDir <- paste(timestampDir, id, sep="/")
```
jiwalker-usgs: you can use `file.path()` to do this rather than paste (I don't know if it is really much better)

R/publish.R
```diff
@@ -84,7 +84,7 @@ publish.resource <- function(viz) {
   # going to start out with simple images
   file <- export(viz)
   dir.create(dirname(file), recursive = TRUE, showWarnings = FALSE)
-  file.copy(viz[['location']], file)
+  file.copy(viz[['location']], file, overwrite = TRUE)
```
jiwalker-usgs: :+1:

R/fetchTimestamp.R
```diff
@@ -78,16 +89,34 @@ fetchTimestamp.file <- function(viz) {
 #' @export
 fetchTimestamp.url <- function(viz) {
   
-  #How will a url be recognized?  Specified in yaml
-  url <- viz$location
-  tag <- headers(HEAD(url))[['last-modified']]
+  #URL will be specified in viz.yaml: will the attribute be called url?
+  url <- viz$url
```
jiwalker-usgs: currently this attribute is `remoteURL` in `fetch.url`

tests/testthat/tests-vizlab.R
```diff
@@ -85,4 +91,29 @@ test_that("getting template from library works", {
...
+  expect_false(fetchTimestamp('siteTextData'))
+  #with a different timestamp
+  writeTimestamp(as.POSIXct("1991-10-21 12:00:00"), 'vizlab/make/timestamps/siteTextData')
+  expect_true(fetchTimestamp('siteTextData'))
+})
+
+test_that(".url works", {
+  #a site with no modified tag
+  viz <- list(id="foo", url="www.google.com")
+  expect_true(fetchTimestamp.url(viz))
```
jiwalker-usgs: rather than call fetchTimestamp.url directly, you could call `as.viz()` and `as.fetcher()` so you can follow the normal execution path in the test.

jiwalker-usgs:
Also, take a look at travis failure.

R/fetchTimestamp.R
aappling-usgs: They say it's faster, which is why I sometimes use it. but I've never seen a case where speed in pasting two strings together into a file path noticeably changed the speed of an application.

wdwatkins:
Suspect there are directory path things happening that don't work on Travis

wdwatkins:
On Windows and Linux there is an extra second being added to the old timestamp, that is only shown when converted to numeric 😒

wdwatkins:
Don't actually need that test, since fetch.timestamp.file doesn't need to do anything

Issue #36, opened 2016-08-11
Accessibility [1]


ldecicco-USGS:
Not sure if vizlab is the place or dinosvg, but would be nice to insert a title by default in all svgs:

https://www.sitepoint.com/tips-accessible-svg/

jread-usgs:
Yeah, it would be `dinosvg` for that
---
Jordan Read, PhD | U.S. Geological Survey | Office of Water Information
608-821-3922 | water • data • science <http://cida.usgs.gov/datascience.html> | g-scholar <https://scholar.google.com/citations?hl=en&user=geFLqWAAAAAJ>

eread-usgs:
We shouldn't violate 508 compliance. 
For svgs:
- [ ]  add title
- [ ]  add text node

lindsaycarr:
Example of adding 508 compliance features for climate-fish-viz:
https://github.com/USGS-VIZLAB/climate-fish-habitat/pull/149
https://github.com/USGS-VIZLAB/climate-fish-habitat/pull/163

SVG: `<object title="">ALT TEXT</object>`
IMG: `<src alt="" title=""></src>`

lindsaycarr:
no script tag is now issue #57; everything else closed in #53

PR #53, opened 2016-09-19
mimetypes lookup dictionary + 508 compliance

lindsaycarr:
Working on #39 and #36


lindsaycarr:
Still need a way for someone to specify a different mimetype config file

lindsaycarr:
@jiwalker-usgs for someone to specify their own dictionary, I am thinking they should need to add a field to their viz.yaml for each section in fetch that is called `mimetypeDictionary` or `mimetypeLookup` or something. Thoughts?

note to self -- would need to add a test to see if that works for the package

jiwalker-usgs:
I think rather than each section defining additional mimetypes, this could live in the root of the viz.yaml as more mimetypes for the particular viz.

jiwalker-usgs:
Also, there is a similar mimetype lookup in the publish section, can we get that to follow this pattern as well.

PR #58, opened 2016-09-19
Ga firstclass WIP

jiwalker-usgs:
Need to implement the injection of analytics into templating


eread-usgs:
per #7

jiwalker-usgs:
@lindsaycarr could you take a look at this?

R/analytics.R
```diff
@@ -0,0 +1,79 @@
...
+#'
+#' @importFrom xml2 read_xml xml_has_attr xml_attr xml_attr<-
+#' @rdname analytics
+#' @export
+analytics.scroll <- function(viz) {
+  required <- c("output")
+  checkRequired(viz, required)
+
+  class <- "vizScroll"
+  html <- read_xml(viz[['output']])
```
jiwalker-usgs: I'm not super happy with the data flow here, but the section publish occurs and adds its output to the `viz` in the output slot.  Then analytics gets called and does its stuff on that output.  Without enforcing a strict pattern for analytics, we end up with this loose arrangement.

R/analytics.R
```diff
@@ -0,0 +1,79 @@
...
+  if (xml_has_attr(html, "class")) {
+    xml_attr(html, "class") <- paste(xml_attr(html, "class"), class)
+  } else {
+    xml_attr(html, "class") <- class
+  }
+  viz[['output']] <- format(html)
```
jiwalker-usgs: `format.xml` prints the content without extraneous information (prolog) which is what I need to eventually write the html out using whisker.

R/publish.R
```diff
@@ -28,11 +28,19 @@ publish.page <- function(viz) {
...
+  # add automatic dependencies
+  dependencies[['_vizlabJS']] <- getVizlabJS()
+
+  # publish all dependencies
+  dependencies <- lapply(dependencies, publish)
 
   context <- buildContext(viz, dependencies)
 
+  #also manually put resources into context
+  context[['resources']] <- append(context[['resources']], dependencies[['_vizlabJS']])
```
jiwalker-usgs: Good call here, I just need to add that id to the viz[['context']] before calling buildContext, I was all turned around with that function (which I wrote :( )

R/publish.R
```diff
@@ -82,10 +93,14 @@ publish.section <- function(viz) {
...
   # figure out resource type and hand to resource handler
   # going to start out with simple images
-  file <- export(viz)
-  dir.create(dirname(file), recursive = TRUE, showWarnings = FALSE)
-  file.copy(viz[['location']], file, overwrite = TRUE)
-  viz[['relpath']] <- relativePath(file)
+  destFile <- export(viz)
+  dir.create(dirname(destFile), recursive = TRUE, showWarnings = FALSE)
+  srcFile = viz[['location']]
+  if (!is.null(viz[['packaging']]) && viz[['packaging']] == "vizlab") {
```
jiwalker-usgs: I was leaving this open for other ways of packaging the files (zips or urls or something), but this is the only reasonable use of it now.

R/publish.R
```diff
@@ -82,10 +93,14 @@ publish.section <- function(viz) {
...
+    srcFile = system.file(srcFile, package = "vizlab")
```
jiwalker-usgs: yeah, not sure how I managed to do that

PR #60, opened 2016-09-20
Vizlab landing page starting code

mwernimont:
@jiwalker-usgs Heres some code, although a lot will change once you get that fancy function going.


inst/landing/index.html
```diff
@@ -0,0 +1,30 @@
...
+
+<body>
+
+<div id="header"></div>
+
+<div id="container">
+
+	<div id="visualizationsContainer"></div>
+	
+	<div id="tagsContainer"></div>
```
jiwalker-usgs: I'm not quite sure how the tags fit in here.

inst/landing/css/.crunch
```diff
@@ -0,0 +1,22 @@
+{
```
mwernimont: Crunch is what I use for LESS compiling. But I don't think we need that exact file

inst/landing/index.html
jiwalker-usgs: Ok, I guess there is a content mismatch there then.  We probably will want to add tags to each visualization, because we don't have that currently.

PR #63, opened 2016-09-21
Remove tag section and Footer template

mwernimont:
@jiwalker-usgs I removed the tags div and css

Also pushed the footer template, although I have sneaking feeling its not what you want, so just let me know what I need to change for it to work.


jiwalker-usgs:
I'm a bit confused about this (the footer, not the tags), I'll talk with you in person about this.

mwernimont:
@jiwalker-usgs Look better?

jiwalker-usgs:
We should be able to work with this.

Issue #29, opened 2016-06-27
switch to safer password storage [1.5]


aappling-usgs:
profile & password info is currently in user/local and depends on .gitignore to never commit that info to the repository. something safer (storing password files outside the repo, for starters) would be great.

eread-usgs:
Just for SB, there is an authenticateSB script, that's in the repo but not checked in. We want to make an external.vizlab folder that can contain this and cross project, not be checked in. With this we should avoid clear text passwords, by encryption. 

Encryption would be nice to have, main priority is to get it out of the repo. 

If we are running this on a dev server with CI, need a way to set the pw into its environment.

wdwatkins:
Are we ok with introducing a new package dependency for encryption?  `digest` seems to be an option.

jiwalker-usgs:
I'd be ok bringing in `digest`

eread-usgs:
- [ ] update  documentation

- [ ] check in the fetch from auth remote file to see if there's an active sb session

wdwatkins:
#64 should finish this.

PR #64, opened 2016-09-21
Password storage 

wdwatkins:
re #29 

Only storing credentials in home/.vizlab now, and authenticating during make with stored creds or with the sbAuthenticate function (not sure if the token from this would still be valid during make?).

Credentials are stored as raw written in .RDS files.

Documentation still needs to be updated, just wanted this looked at first


R/authRemote.R
```diff
@@ -47,11 +42,36 @@ authRemote.default <- function(fetcher, user='local', ...) {
 #' @rdname authRemote
 #' @export
 authRemote.sciencebase <- function(fetcher, user, ...) {
-  auth.script <- paste0("user/", user, "/auth.sciencebase.R")
-  if(file.exists(auth.script)) {
-    source(auth.script)
+  #change to home directory for storage
+  home <- path.expand('~')
```
jiwalker-usgs: Does this work on Windows? I think `Sys.getenv("HOME")` might be more cross platform

R/authRemote.R
```diff
@@ -47,11 +42,36 @@ authRemote.default <- function(fetcher, user='local', ...) {
...
+  dotVizlab <- paste(path.expand("~"),".vizlab", sep="/")
+  if(!dir.exists(dotVizlab)){
+    dir.create(dotVizlab)
+  }
+  sbCred1 <- readline(prompt = "Enter username: ")
+  sbCred2 <- readline(prompt = "Enter password: ")
+  sbtools::authenticate_sb(sbCred1, sbCred2)
+  storeCreds <- readline(prompt = "Store credentials? (T/F): ")
+  if(storeCreds){
+    saveRDS(charToRaw(sbCred1), paste0(dotVizlab, "/cred1"))
```
jiwalker-usgs: This file name is different than above

R/authRemote.R
```diff
@@ -47,11 +42,36 @@ authRemote.default <- function(fetcher, user='local', ...) {
 #' @rdname authRemote
 #' @export
 authRemote.sciencebase <- function(fetcher, user, ...) {
-  auth.script <- paste0("user/", user, "/auth.sciencebase.R")
-  if(file.exists(auth.script)) {
-    source(auth.script)
+  #change to home directory for storage
+  home <- path.expand('~')
+  sbCred1 <- paste0(home, "/.vizlab/sbCred1")
```
jiwalker-usgs: I think it would be better to store as one file, sbCred1 and sbCred2 seems like we are storing two different credentials rather than two parts of one.

R/authRemote.R
```diff
@@ -47,11 +42,36 @@ authRemote.default <- function(fetcher, user='local', ...) {
...
+  sbCred2 <- paste0(home, "/.vizlab/sbCred2")
```
jread-usgs: is `file.path()` better than `past0` for these types of things?

wdwatkins: It puts the slashes in for you at least.

R/authRemote.R
jiwalker-usgs: Could you make a list of raws and save that to RDS
```r
creds <- list(username=charToRaw(sbCred1),password=charToRaw(sbCred2))
saveRDS(creds, file.path(dotvizlab, "sbcreds"))
```

R/authRemote.R
wdwatkins: It worked for @lindsaycarr.  `HOME` could be better though.

jiwalker-usgs: Shows how much I know about Windows :dizzy_face:

PR #75, opened 2016-10-06
CSS Fixes

mwernimont:
@jiwalker-usgs What did I do wrong?


jiwalker-usgs:
We don't want to check in this folder, this is copied from inst for the process of building

mwernimont:
@jiwalker-usgs Look better?

jiwalker-usgs:
beautiful!

Issue #42, opened 2016-08-23
Documentation- Vignette [3]


eread-usgs:
Create a vignette and complete documentation for the vizlab framework. 

Describe the steps to create a new viz using the package, describe in more detail the viz yaml contents and the purpose, a couple of examples, pull from the climate fish viz as examples. How to run the make. Steps to iterate on a viz. 

Fill out example repo as the psuedo-vignette, a complete but simple implementation of the framework.

eread-usgs:
Check out sbtools or the 2nd geoknife vignette as examples of right-sized documents
This should cover ~ 80% of the use case. Narrative for steps to creating and publishing a rapid viz.

lindsaycarr:
This should get pulled out from README and put into this doc: https://github.com/USGS-VIZLAB/vizlab/pull/53/commits/2b4d263fde5ba27d7eec1660c206efba154eee94

eread-usgs:
Should include section on profile.yaml (per #3) including instructions for how to set up the profile for first time users.

lindsaycarr:
- [ ] overview: what can this package be used for?
- [ ] make workflow: what is make and why is it useful?!
- [ ] setup instructions: vizSkeleton, createProfile, createMakefiles
- [ ] sb authentication **_where should this go in the order?_**
- [ ] what is a viz.yaml?
- [ ] workflow: fetch, read, process, visualize, publish
- [ ] using make
- [ ] example: putting it all together

wdwatkins:
I was thinking just go through the example at the end of each step.

wdwatkins:
#67 work in progress

wdwatkins:
#70 is mostly complete

wdwatkins:
Merged #70, vignette should be finished.  Should be updated though if the example viz is updated, especially the viz.yaml

wdwatkins:
Needs updated to include footer and other new additions

eread-usgs:
Include the viz objects how they get passed into functions.

eread-usgs:
build this into package so it can be found more easily by users.

eread-usgs:
During Matthew, needed quick way to find out to deal w/ dependencies and couldnt find it.

wdwatkins:
Now:
[] viz object passed into all functions
[] footer

pending other issues:
[] multiple dependency management

eread-usgs:
Needs review following recent package modifications.

PR #99, opened 2016-10-18
vignette setup

wdwatkins:
#42 .Rmarkdown is actually accessible as vignette.  Still needs updating


DESCRIPTION
```diff
@@ -29,5 +30,6 @@ Suggests:
     knitr,
     readxl,
     sbtools (>= 0.19.3),
-    testthat
+    testthat,
+    rmarkdown
```
jiwalker-usgs: I've been trying to keep Imports/Suggest in alphabetical order (read that as a recommendation somewhere).  But don't worry about this, I'll probably be moving it up to Imports with my markdown PR.

Issue #47, opened 2016-09-08
support markdown syntax for links [2]


eread-usgs:
Links for starters, there may be a couple other things in addition to this.

eread-usgs:
use markdown package?

jiwalker-usgs:
The pattern for this might still be a bit wonky at this point, but there is support for markdown at this point.

PR #100, opened 2016-10-18
Markdown

jiwalker-usgs:
I ended up messing around with the page and section publishers for handling dependencies.  This will hopefully not affect anything adversely, but I can see that this is getting a bit unruly.  Do you see any issues or improvements to this approach @lindsaycarr @wdwatkins?


jiwalker-usgs:
fix for #47

R/export.R
```diff
@@ -73,6 +75,19 @@ export.css <- function(viz) {
   return(file)
 }
 
+#' Default export, defaults to not export
```
wdwatkins: I'm a bit confused by this description

jiwalker-usgs: This is the default export method.  It defaults to not export the contents to the target directory and requires either an override or explicitly setting `export: true` on the object.

R/publish.R
```diff
@@ -65,8 +70,18 @@ publish.section <- function(viz) {
   checkRequired(viz, required)
 
   # TODO Watch out for cyclic depends
-  dependencies <- lapply(viz[['depends']], publish)
+  dependencies <- as.list(viz[['depends']])
+  dependencies <- lapply(dependencies, function(x) {
+    expanded.dep <- NULL
```
wdwatkins: Looks like you could break this out as a named function since it's also used above?

wdwatkins:
Looks ok to me, so the main difference is that the output of all the published dependencies is being returned inside the `lapply`?  In addition to the `cat` at the end?

R/publish.R
jiwalker-usgs: good idea, I thought these might be different, but they ended up equivilent.

jiwalker-usgs:
The main difference is that it is doing publish and/or readData so it can pick up on dependencies that are nested.  It also introduces the dot notation for nested dependencies, which is hopefully easy to follow.  Added to example https://github.com/USGS-VIZLAB/example/pull/12.

lindsaycarr:
:100: !!!!

PR #101, opened 2016-10-20
Fix for bug introduced in markdown logic

jiwalker-usgs:


jiwalker-usgs:
Successfully built https://github.com/USGS-VIZLAB/hurricane-matthew as a test for this fix.

Issue #91, opened 2016-10-11
Manage viz R packages required (1)


jiwalker-usgs:
The setup to get a viz to build requires all packages being in place.  There should be a section in viz.yaml or somewhere else reasonable to define R install details (CRAN, GRAN, github, ...).  Also, these need to be loaded so we can remove `library()` calls from viz scripts.

eread-usgs:
For setup, need to figure out what is needed to install. If VizLab package is modified, then there's no way to indicate if github versus another version is needed. 

make as a section in the viz.yaml.

eread-usgs:
1) Where do we define our dependencies?
2) How do we install?
3) How do we load?

eread-usgs:
do '= version number' not '> version number '

eread-usgs:
use rOxygen to create a namespace file?

eread-usgs:
add libraries/packages section to viz skeleton

aappling-usgs:
We have accomplished the first part of this task "section in viz.yaml or somewhere else reasonable to define R install details (CRAN, GRAN, github, ...)".

We've added `checkVizlabPackages()` to help with the second part ("How do we install?"), though it's not a complete solution because it doesn't do any automatic installation.

Remake (#246) would be one way to address the third part ("How do we load?"), though if we stick with make we'll need some other solution.

aappling-usgs:
Closing this issue because all tasks have been completed or have been given their own issues

PR #106, opened 2016-11-18
required-packages in viz.yaml

lindsaycarr:
vizSkeleton now creates an element called `required-packages` that will be a collection of the packages needed to run this visualization. `vizlab` is automatically added as the first package.

e.g. `required-packages: vizlab, gsplot, dataRetrieval`

fixing part of #91


wdwatkins:
Wouldn't you have to have vizlab installed already to run this function?

lindsaycarr:
yes, this is just a section stating what packages the viz uses to users. It doesn't install or load anything. Just for documentation

wdwatkins:
Ok.  It seems like we might want to have a function that loads packages based on this, but it could just know to skip vizlab then.  We probably want to have some system to provide corresponding repos for each package, at least if it isn't CRAN

lindsaycarr:
Yes, but we decided per #91, that we would start with just a place to define things since we couldn't come to consensus about how to load.

wdwatkins:
Sorry, I just meant the loading function could be a possible feature that would utilize this list, not that you should add it right now.  But for listing things, the repo and/or version could be important to know also.   This is fine to start with though.

lindsaycarr:
ah now I see what you mean. Hmmm...let me think about how I could add that.

lindsaycarr:
Ok, here's an idea for including different repositories. Any ideas about how to get the version on there @wdwatkins ?

```
required-packages:
    cran: dataRetrieval
    gran: vizlab
    github: USGS-R/gsplot
```

wdwatkins:
I was thinking something like that, but then the only way I can think of would be to have a corresponding versions vector, which seems pretty clunky:

```
required-packages:
    cran: dataRetrieval
         versions: xxx 
    gran: vizlab
         versions: xxx
```

Another way might be to make each package its own thing, but that would make the viz.yaml get really long:

```
required-packages:
        dataRetrieval:
                repo: GRAN
                version: xxx
       github:
                 repo: github
                 version:  yyy
```

lindsaycarr:
Let's see which @jiwalker-usgs likes best when he gets back (not a difficult change to merge the day before we kick off the viz)

jiwalker-usgs:
I think I like the last option best:
```
required-packages:
  dataRetrieval:
    repo: GRAN
    version: xxx
  github:
    repo: github
    name: cscheid/rgithub
    version: yyy
```

Though repo GRAN is ambiguous in this case.  We'd want to define the repo options before attempting to automate anything out of this.

jread-usgs:
use `tag` instead of `version` for github?

aappling-usgs:
Yeah, or `ref` because that's what `install_github` uses? https://www.rdocumentation.org/packages/devtools/versions/1.12.0/topics/install_github

jread-usgs:
yeah, that is better. the version is ambiguous because `intall_github` is based on tags or commits, and you can tag `1.0.0` when the actual version of the package is totally different. `ref` allows us to cover tags and commits and be consistent there.

jiwalker-usgs:
I propose that we work through this section in the new viz and backport what we come up with into the skeleton.  I also think that calling out vizlab in this section might be weird to @wdwatkins point, so we should discuss that further.

jiwalker-usgs:
I'm going to close this for now, we can revive it after this viz and a discussion on the pattern we want going forward.

aappling-usgs:
Oh, I see why these changes got lost - they never got merged! I will add them again as I'm working through  #123

lindsaycarr:
Ahhh yes, that would do it :)

Issue #81, opened 2016-10-07
profile.yaml in wrong place from createProfile() on windows (1)


aappling-usgs:
`createProfile()` puts `profile.yaml` in `C:\Users\aappling\Documents\.vizlab`, but my shell looks for the profile in `C:\Users\aappling\.vizlab`. It's the difference between RStudio's and make's idea of `~`. I had to move it manually to get `make` to run.

aappling-usgs:
also, once i've moved it to where it needs to be for `make`, i get
```r
> createMakefiles()
Error in findProfileYaml() : 
  profile.yaml does not exist in a supported directory. See createProfile().
```

eread-usgs:
Make sure the solution is supported on mac, linux and windows.

aappling-usgs:
I think the truly optimal approach would be to become smarter about the differing values of $HOME in RStudio and bash, and to use that info to create and/or search for this file in the right place. But I've tried and don't see how to do that cleanly, so both #107 and #108 take the more expedient approach of giving the user control in `createProfile()` over where profile.yaml gets written. Probably fine for the next year or two.

PR #107, opened 2016-11-19
gave user the ability to give specific filepath for profile.yaml and …

lindsaycarr:
…updated documentation

fixing #81

### DIFF
--- a/R/profile.yaml.R
+++ b/R/profile.yaml.R
@@ -3,10 +3,16 @@
 #' This creates a file that describes where necessary executables are located in order to run make.
 #' This sets up a template, it is necessary to fill in the missing information
 #'
-#' @param directory location to save the profile.yaml file. There are currently 2 options: "home" or "relative"
+#' @param directory location to save the profile.yaml file. There are currently 3 options: "home" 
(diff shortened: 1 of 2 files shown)


aappling-usgs:
What's the rationale for this: https://github.com/lindsaycarr/vizlab/commit/7be7f1236b5b8274631c601bf15ab33d26976c32?

aappling-usgs:
I ask about  lindsaycarr@7be7f12 because this approach limits the user's flexibility: even though your changes let me create `vizlab/profile.yaml` anywhere I want, I still can't create `.vizlab/profile.yaml` anywhere I want. The latter is the use case I had in mind when I created #81, though I wonder if a more flexible `findProfileYaml()` would be better than a flexible `createProfile()`. I get that a flexible `createProfile()` is easier to implement and might make sense for now.

I'm going to propose a different PR that competes/conflicts with this one. It follows the same approach of modifying `createProfile()` rather than `findProfileYaml()`. The major difference is that the new PR will accept a file path directly, with a default of `~/.vizlab` and notes in the parameter description about other valid options.

lindsaycarr:
Closing because #108 fixes this instead.

PR #108, opened 2016-11-19
retry on #81

### LINKED ISSUE #81: profile.yaml in wrong place from createProfile() on windows (1)

aappling-usgs:
@lindsaycarr - review?

This PR competes with your #107. I'm proposing that we close #107 and merge this one instead.

This PR follows the same approach as #107 of modifying `createProfile()` rather than `findProfileYaml()`. The major difference is that this PR will accept a file path directly, with a default of `~/.vizlab` and notes in the parameter description about other valid options. This offers the flexibility I think is needed to resolve #81.

I let some clutter into this PR, sorry. There are some minor edits in other files. The essential changes are all in the block of red and green that starts here: https://github.com/USGS-VIZLAB/vizlab/commit/51ec4544ab704cb92aa7c5743cd121345153257e#diff-020451bb865d7e219edcb3e30c710de4L22, plus some changes in that file to reduce code redundancy.


lindsaycarr:
@aappling-usgs this looks good to me! Could you update the documentation for `findProfileYaml`? It still references the home or relative options for `createProfileYaml`.

aappling-usgs:
Yep, good idea!

PR #111, opened 2016-11-23
Adding favicon

ldecicco-USGS:
Not sure the img folder will properly build into the target. I'm trying now to figure out how to test that.


ldecicco-USGS:
OK someone can review this and merge if it seems to work.

jiwalker-usgs:
I don't think this will exactly work.  There needs to be an explicit publish of the favicon.ico in order to move it from the images directory to the target/images in the publish phase.  I think this is bigger than a favicon task though, as we need all the components needed by a page to be explicitly called out to the publish process.

ldecicco-USGS:
I assumed that was happening somewhere in here:
https://github.com/USGS-VIZLAB/vizlab/blob/master/R/publish.R#L120

It's not?

jiwalker-usgs:
Yes, that is where this happens, but the resource (img in this case) needs to explicitly called out in the dependency chain.  The file existing in that directory is not enough.

jiwalker-usgs:
My comment https://github.com/USGS-VIZLAB/vizlab/blob/master/R/publish.R#L32 relates to the need to bring in automatic dependencies such as this.  It is only done for the "special" js file right now, but a more general solution is needed.

ldecicco-USGS:
Ohhhh, OK.

R/publish.R
```diff
@@ -122,6 +122,10 @@ publish.img <- function(viz) {
   viz <- NextMethod()
   checkRequired(viz, required)
 
+  index_fav <- 'target/img'
+  if(!dir.exists(index_fav)) dir.create(index_fav, recursive=TRUE)
+  file.copy(from=system.file('img/favicon.ico', package="vizlab"), to=index_fav)
+
```
ldecicco-USGS: @jiwalker-usgs something like this?

jiwalker-usgs: Not so much, the favicon needs to have an object created to represent it.  The copy will then occur because it is of type "resource" and will go to the right place because of the mimetype lookup.  This object needs to be added as a resource on the "page" type, which will automatically publish it.  It is more complex than I had thought through, I'll probably try to work this in with our current viz.

Issue #92, opened 2016-10-11
Using dependencies from viz.yaml (2)


jiwalker-usgs:
The `viz[['depends']]` needs a bit of love to make it work for the script pattern we have.  Currently there are two ways to access something from depends:

``` r
process.timeAndMap <- function(viz) {
  timeData <- readData(viz[['depends']][1])
  mapData <- readData(viz[['depends']][2])
}
```

or if yaml looks like

``` yaml
-
  id: myProcess
  processor: timeAndMap
  depends: [ "firstDependency", "myDependency" ]
  timeData: firstDependency
  mapData: myDependency
```

then script can do something like:

``` r
process.timeAndMap <- function(viz) {
  timeData <- readData(viz[['timeData']])
  mapData <- readData(viz[['mapData']])
}
```

neither are very satisfying, as well as needing to call readData from script rather than having it at the ready.  So a couple topics of discussion:
1. Should we switch depends to a named list so code ends up looking like second script without duplication of ids?
2. Should data be read in automatically so script does not need to call `readData()` directly? (I don't think this is the main issue, but worth discussing)
3. Do we want to formalize in some way the second approach that the depends section isn't even required? (I'm thinking here maybe we replace all references with a sigil that denotes a dependency so we can still build our dependency graph; this would look something like `@id`, `&id`, or `!id`).
4. Are there other things here that I'm missing, what is the natural approach for using dependencies?

jread-usgs:
Maybe a bad idea (although not sure why), but what about keeping the dependencies with the code somehow and leaving them out of the yaml, kind of like `#' @importFrom pkg fnc`

something like 
```
myfun <- function(viz){
    depends("firstDependency", "myDependency")
    time.range <- range(firstDependency)
    return(time.range)
}

**Comment by aappling-usgs** [comment]
Agreed, @jread-usgs - I like the idea of keeping dependencies with the code wherever possible.

Something that I haven't fully understood is why the `viz[['depends']][1]` paradigm rose to the top. Would you do it differently if I told you that
```r
pd <- readData('precip-data')
```
also works? When I was sketching out pieces of vizlab this summer I had something more like this in mind:
```r
processData.mungePrecip <- function(viz, precip_data=readData('precip-data')) {
  ...code using precip_data...
}
```

you still currently have to copy the dependencies from the function argument list over to viz.yaml (maybe that could be fixed) but even now I think I'd find it easier to use names in both places rather than using dependency indices within the function.

[...]

PR #112, opened 2016-11-23
named depends in testviz viz.yaml

wdwatkins:


wdwatkins:
An approach to #92

wdwatkins:
@jiwalker-usgs @jread-usgs @lindsaycarr If we take this approach maybe we should have a naming convention for the different depends items?

lindsaycarr:
Wouldn't they just be the id of the thing it's referring to?

wdwatkins:
That would be one option, although we could conceivably want to use something more general in case we changed what the dependencies were (i.e. a generic 'data' dependency or something).

jiwalker-usgs:
I think the names should be descriptive of what they are used for, and a convention is a good idea in general so we don't have to second guess ourselves in looking back.  I've avoided using CAPITAL letters in the yaml keys, though that is just one possibility.

PR #116, opened 2016-11-30
should fix #115

### LINKED ISSUE #115: make fails unless make clean is run
call function script doesnt get copied over unless you do `make clean`.

aappling-usgs:
@jiwalker-usgs - review? i think this should fix the issue we just identified with `callFunction.R` not always getting copied over.

### DIFF
--- a/R/createMakefiles.R
+++ b/R/createMakefiles.R
@@ -173,12 +173,12 @@ createMakeDirs <- function(makefile) {
     if(!dir.exists(timestampdir)) dir.create(timestampdir)
   }
 
-  # if any rules use callFunction.R, copy that script into the vizlab/make
-  # directory. this script allows us to call a single function via R CMD BATCH
-  if(any(grepl('callFunction.R', makefile))) {
-    file.copy(system.file('scripts/callFunction.R', package='vizlab'),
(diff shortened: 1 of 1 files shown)


jiwalker-usgs:
I think this should do it, but I'm not sure why it wasn't working.  Does it maybe need `recursive=TRUE` on the `file.copy`?  It could be that the `vizlab/make` directory doesn't exist at that point?

aappling-usgs:
My guess is that it wasn't working because when other people today called `createMakefiles`, they didn't yet have items in viz.yaml that required making, so `callFunction.R` didn't get added to any makefiles, so the file didn't get copied (because it was in that `if()` block). Compare this to my experience today, pulling other people's additions to viz.yaml before i ever called createMakefiles: the file got copied for me, no problem, and I never had to call `make clean`.

jiwalker-usgs:
I think that is right, that would be my situation at least.

aappling-usgs:
i think the order of function calls means `vizlab/make` should exist before that `file.copy` line gets called. here are the key calls, in order, in createMakefiles:
```r
createMakefiles <- function() {
  createTopMakefiles() # vizlab/make directory gets created here
  createBlockMakefile() # calls createMakeDirs which calls file.copy
}
```

PR #118, opened 2016-12-02
handle relative + absolute related viz paths

lindsaycarr:
@jiwalker-usgs


jiwalker-usgs:
We'll have to look at the test failures.

PR #119, opened 2016-12-03
support Ico

jiwalker-usgs:
Also snuck in the finish for adding data resources (stop message was preventing it)


jiwalker-usgs:
Merging this, the test failure is a fragile publishLanding test

PR #128, opened 2016-12-10
don't override name of viz if set elsewhere

jread-usgs:


jread-usgs:
@jiwalker-usgs can you review? not sure if you had a different pattern in mind

jiwalker-usgs:
Yeah, I think this is fine.  I'm not sure we were expecting names coming from this side of things, but it seems like a reasonable thing to override.

PR #130, opened 2016-12-13
use thumbnail-landing for landing page

lindsaycarr:
still need to run `publishLandingPage()` when all the viz.yamls get updated to check that this works


jiwalker-usgs:
Can you update tests?

PR #132, opened 2016-12-14
thumbnails will work now??

lindsaycarr:
@jiwalker-usgs thumbnails are still broken, I think this fixes it. It should be relative to vizlab, so it needs the correct viz before the thumbnail path. Can't really test locally though, so hopefully I can see this on dev after I merge and it's deployed.

### DIFF
--- a/R/publishLandingPage.R
+++ b/R/publishLandingPage.R
@@ -66,8 +66,9 @@ getVizInfo <- function(org, repo){
                             template="templates/vizzies.mustache",
                             publisher="section",
                             context=list(name=viz_info$name,
-                                         thumbnail=file.path(getVizUrl(viz_info$path),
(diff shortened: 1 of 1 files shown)


R/publishLandingPage.R
```diff
@@ -66,7 +66,8 @@ getVizInfo <- function(org, repo){
                             template="templates/vizzies.mustache",
                             publisher="section",
                             context=list(name=viz_info$name,
-                                         thumbnail=getVizUrl(viz_info$`thumbnail-landing`$url),
+                                         thumbnail=file.path(getVizUrl(viz_info$path),
```
jiwalker-usgs: right, though file.path might not be the right thing to use here

lindsaycarr: ok, just paste with '/' then

PR #140, opened 2017-01-07
fix typo in vignette (install.github -> install_github)

aappling-usgs:
Typo pointed out by a potential user - thanks, Jonathan!


jiwalker-usgs:
We should be able to start directing people to GRAN as well.

PR #147, opened 2017-02-03
vizlab to match water use footer

mwernimont:
@jiwalker-usgs Took the template from water use and put it in the inst footer template, and then switched the vizlab footer CSS for the water use footer CSS.


jiwalker-usgs:
I think this dropped the related blog posts that were optional in the old format.

mwernimont:
@jiwalker-usgs Look better?

jiwalker-usgs:
Close, because the blogs are optional there was a boolean that enabled that section.

mwernimont:
@jiwalker-usgs How's that? Want me to remove that other file from the PR?

Issue #141, opened 2017-01-10
Thumbnails - research (1)


eread-usgs:
- [ ] Specify standard sizes (FB, Twitter, our landing page)

- [ ] Get Thumbnail for our landing page

- [ ] OTher issues?

eread-usgs:
We need a target thumbnail size b/c twitter , FB have different rules. Force compliance on size.

Consider: Make sure landing page thumbnail sizes are compatible.

eread-usgs:
Check file size and pixel dimensions - max and min?

wdwatkins:
https://developers.facebook.com/docs/sharing/best-practices

> Use images that are at least 1200 x 630 pixels for the best display on high resolution devices. At the minimum, you should use images that are 600 x 315 pixels to display link page posts with larger images. Images can be up to 8MB in size.

https://dev.twitter.com/cards/types/summary-large-image
> Images for this Card should be at least 280px in width, and at least 150px in height. Image must be less than 1MB in size.

wdwatkins:
biggest image that would fit twitter is a little under 600x600 px, which isn't ideal for facebook.

Read about pixels here: http://shutha.org/node/796

wdwatkins:
@jiwalker-usgs should a publisher for this do anything with the cards in the html?  Or just check the dimensions?

jiwalker-usgs:
I think the publisher will mostly be moving files around and deciding on naming, but it would be useful to include "compliance" as part of that to verify that images meet the specs.

PR #153, opened 2017-02-23
start thumb compliance

wdwatkins:
#141


inst/testviz/viz.yaml
```diff
@@ -136,6 +136,12 @@ publish:
     depends: relativeAbundanceFig
     publisher: section
   -
+    id: facebook-thumb
+    location: images/facebook-thumb.png
+    publisher: thumbnail
+    for: facebook
+    mimetype: png
```
wdwatkins: Does there need to be a specific png mimetype, since this file only needs to be moved?

jiwalker-usgs: the mimetype determines the target folder (css, js, images, data) so setting this to "image/png" or "image/jpg" will be needed to get it to the right place.

R/publish.R
```diff
@@ -282,6 +280,53 @@ publish.landing <- function(viz){
   publish(pageviz)
 }
 
+#' check dimensions and size, publish thumbnail
+#'
+#' @rdname publish
+#' @export
+publish.thumbnail <- function(viz){
```
jiwalker-usgs: Thumbnail should be a sub-class of resource so that the publish call gets dispatched here like a normal image, and then after the checks pass, you should be able to call `NextMethod` to progress up the class hierarchy which should be something like `class(viz) = c("thumbnail", "img", "resource", "publisher", "viz")`.  Thoughts on that?

R/publish.R
```diff
@@ -282,6 +280,53 @@ publish.landing <- function(viz){
...
+  checkThumbCompliance(file = viz[['location']], maxSize = maxSize,
+                       minHeight = minHeight, minWidth = minWidth, square = square)
+  #send to resources publisher if all ok
+  viz <- as.resource(viz)
+  publish.resource(viz)
+  
+}
+
+#helper to check thumbnail compliance
+checkThumbCompliance <- function(file, maxSize, minHeight, minWidth, square = FALSE) {
```
jiwalker-usgs: This is pretty handy.  It is probably best to do `@importFrom` for imager and add it to the DESCRIPTION file.

wdwatkins:
The .Rd files all changed because of the Roxygen update

wdwatkins:
@jiwalker-usgs can you review my changes?

R/publish.R
```diff
@@ -282,16 +280,71 @@ publish.landing <- function(viz){
...
+  } else { #landing
+    maxSize <- 1048576
+    checkHeight <- 400
+    checkWidth <- 400
+  }
+  dims <- checkThumbCompliance(file = viz[['location']], maxSize = maxSize,
+                       checkHeight = checkHeight, checkWidth = checkWidth)
+  #send to other publishers if all ok
+  viz <- NextMethod()
+  viz[['url']] <- paste(getVizURL(), viz[['relpath']])#need to add slash between?
```
jiwalker-usgs: This will put a space between them, it may need to add the slash if relpath doesn't contain it.

R/publish.R
```diff
@@ -282,16 +280,71 @@ publish.landing <- function(viz){
...
+  viz[['width']] <- dims[['width']]
+  viz[['height']] <- dims[['height']]
+}
+
+getVizURL <- function() {
+  return(print(getBlocks("info")$info$path[[1]]))
```
jiwalker-usgs: This function should also attach the other information to get the full url (https://owi.usgs.gov/vizlab)

R/publish.R
wdwatkins: thumbnails should get exported right?  `relpath` is currently coming back as null for them

jiwalker-usgs: yes, this shouldn't be null if it is working right, it should be something like "images/thumb1.png"

wdwatkins:
The thumb should end up in `target/images` right?

jiwalker-usgs:
yes, though that might have to be part of the publisher if that needs to happen.

jiwalker-usgs:
Also, sorry about the merge conflicts, I merged the googlefont PR.

wdwatkins:
Hmm yeah, the export sends it to `target/data`

jiwalker-usgs:
Right, we pulled the `img` out of the class list, so the export call hits resource (which defaults to data).  You can implement `export.thumbnail` to map these to the images directory.

wdwatkins:
If the base vizlab url is in the environment, do we need a `AAA.R` like dataRetrieval to set it?

Issue #144, opened 2017-01-10
Common tasks for each viz- create issues for these at the outset of each new viz (2)


eread-usgs:
Use github API to automate creation of these issues?

- [ ] Social media checklist
- [ ] Reviewer checklist
- [ ] Thumbnail
- [ ] Verify thumbnail dimensions and resolution rules through developer documentation
- [ ] Related content
- [ ]  Press release

eread-usgs:
Make an init that is part of the skeleton- only gets run once.

lindsaycarr:
@eread-usgs what is the checklist for reviewers?

lindsaycarr:
@eread-usgs @jiwalker-usgs also wondering what labels would be good to have? outreach, publisher, visualizer, reader, fetcher, UI ?

PR #154, opened 2017-02-23
automatically generate issues w/ labels for a viz

lindsaycarr:
`initializeVizRepo(repo_name="myviz", description="This is a description of my viz")` will create a new repo  within USGS-VIZLAB with 6 new issues with corresponding labels.

Tests for this are commented out (and will remain that way) because there is interactive authentication, plus we don't want to create and delete a repository every time the tests are run. The commented out tests will remain there for us to run manually if we need to.

See #144.


R/initializeVizRepo.R
```diff
@@ -0,0 +1,103 @@
...
+#' 
+#' @return URL for the new GitHub repository.
+#' @export
+initializeVizRepo <- function(repo_name, description, org="USGS-VIZLAB",
+                              issue_json=file.path("inst", "issuetemplates.json"),
+                              label_json=file.path("inst", "labeltemplates.json")){
+  
+  # log in to GitHub & scope to read/write access for repos using profile.yaml
+  # for additional scope info, see https://developer.github.com/v3/oauth/#scopes
+  github_creds <- getProfileInfo()
```
jiwalker-usgs: may want to verify that creds are not null here for a warning with helpful hint, error from github will likely not help too much in telling users to edit their profile.yaml

R/initializeVizRepo.R
```diff
@@ -0,0 +1,103 @@
...
+#' @param repo_name string, name for the new repository
+#' @param description string, short description of the repository
+#' @param org string, GitHub organization to create repository. Defaults to "USGS-VIZLAB"
+#' @param issue_json file path indicating the JSON file to be used to define what issues to create
+#' @param label_json file path indicating the JSON file to be used to define what labels to create
+#' 
+#' @return URL for the new GitHub repository.
+#' @export
+initializeVizRepo <- function(repo_name, description, org="USGS-VIZLAB",
+                              issue_json=file.path("inst", "issuetemplates.json"),
```
jiwalker-usgs: I think the defaults are left from testing, wasn't this supposed to be system.file?

lindsaycarr: whoops - correct! I updated the test I was working on but not this function.

R/initializeVizRepo.R
lindsaycarr: Good idea

eread-usgs:
Sorry for the delayed review- the list looks good to me @lindsaycarr

PR #164, opened 2017-03-08
New simple footer

mwernimont:
@jiwalker-usgs want to give it a look?


inst/templates/footer.mustache
```diff
@@ -1,29 +1,43 @@
...
+      <a href="mailto:vizlab@usgs.gov">
+        <p>Contact Us</p>
+      </a>
+    </div>
+    
     <div id="relatedVizzies">
-    	<h3>Related Visualizations:</h3>
-
-    	{{#vizzies}}
+      <!--{{#vizzies}}-->
```
jiwalker-usgs: What is happening here?

inst/resource.library.yaml
```diff
@@ -160,6 +161,11 @@
   mimetype: text/javascript
   publisher: resource
 -
+  id: lib-header-css
```
jiwalker-usgs: Is there a header template that uses this?

mwernimont:
@jiwalker-usgs The  was a forgotten attempt at getting the github blocking to stop, but didn't see it since mustache files have no code coloring. Removed the .crunch file too. For the header I don't know, do we have a header template that could use it?

inst/templates/footer.mustache
```diff
@@ -1,29 +1,43 @@
...
-    					<img src="{{thumbLoc}}" alt=""/>
-    				</div>
     			</div>
     		</a>
     	{{/vizzies}}
     </div>
-  
-  {{#blogsInFooter}}
+    
+   <!-- {{#blogsInFooter}}
```
jiwalker-usgs: You also commented out blogsInFooter.

mwernimont: Yep, Emily wasn't sure what we are doing with blogs in the footer, so figured commenting it out was the best solution for the moment

jiwalker-usgs: Not a huge fan of commented out code, but ok for now.

PR #168, opened 2017-03-10
Dynamic hover

mbucknell:
Hasn't been tested, but it's basically the same code that was in the water-use viz reorganized.


inst/js/svg.js
```diff
@@ -0,0 +1,79 @@
...
+      var textLength;
+      var halfLength;
+      var tooltipX
+      
+      tooltip.firstChild.data = text;
+      textLength = Math.round(tooltip.getComputedTextLength());
+      halfLength = textLength / 2;
+      
+      /* Make sure tooltip text is within the SVG */
+      if (svgPoint.x - halfLength - 6 < 0)  {
```
jread-usgs: I think `6` is the left buffer between the text and the edge of the tip box. Can that be a variable?

inst/js/svg.js
```diff
@@ -0,0 +1,79 @@
...
+        tooltipX = svgPoint.x;
+      }
+      tooltip.setAttribute("x", tooltipX);
+      tooltip.setAttribute("y", svgPoint.y);
+      tooltip.setAttribute("class", "shown");
+      
+      /* Set attributes for background box */
+      tooltipBox.setAttribute("x", tooltipX - halfLength - 6);
+      tooltipBox.setAttribute("y", svgPoint - 35);
+      tooltipBox.setAttribute("width", textLength + 12);
```
jread-usgs: this is 2*buffer

inst/js/svg.js
```diff
@@ -0,0 +1,79 @@
...
+      else 
+        tooltipX = svgPoint.x;
+      }
+      tooltip.setAttribute("x", tooltipX);
+      tooltip.setAttribute("y", svgPoint.y);
+      tooltip.setAttribute("class", "shown");
+      
+      /* Set attributes for background box */
+      tooltipBox.setAttribute("x", tooltipX - halfLength - 6);
+      tooltipBox.setAttribute("y", svgPoint - 35);
```
jread-usgs: the `35` here is the height of the box, which we might want to vary depending to the font size of the text. Should we parameterize it or have it respond to the text height in some way (if that is possible?)

inst/js/svg.js
mbucknell: The answer is yes. Your idea of adding the code to insert the html for the mouseover text will complicate that a bit since that six comes from there. I think I'll get it working without parameterizing the dimensions and then look into doing that.

PR #172, opened 2017-03-22
fixes to header and footer

jiwalker-usgs:
template changes caused some issues with the header and footer


mwernimont:
It was the USGS header footer templates causing the secondary vizlab header?

jiwalker-usgs:
No, it was the fullpage template where we pulled out header and footer from sections.

PR #173, opened 2017-04-17
A couple fixes for current viz

jiwalker-usgs:


jiwalker-usgs:
Merging to fix issue with analytics sprint

PR #182, opened 2017-09-08
add backslashes to default windows profile.yaml; fixes #181

### LINKED ISSUE #181: createProfile.Windows should add backslash in spaces
e.g., between `Program` and `Files` in 
```
RSCRIPT: C:/Program\ Files/R/R-3.4.1/bin/x64/Rscript.exe
```

aappling-usgs:
here you go, @lindsaycarr - could you try out the new `createProfile`?


lindsaycarr:
Yep - the `\` works, but my path doesn't work with `x64` in there. If I remove it, it runs just fine. Sounds like a different issue though

aappling-usgs:
oh! didn't know about the x64 thing. i'll add that as a separate issue for us to take on when you're not deep in Irmaland.

PR #199, opened 2017-09-11
change semantics

wdwatkins:


wdwatkins:
@ldecicco-USGS is there an example for facebook?  we could put that in too.

wdwatkins:
just gonna change in the viz for now

inst/templates/semantics.mustache
```diff
@@ -9,11 +9,11 @@
   <meta property="og:locale" content="en_US" />
 
   <meta name="twitter:card" content="summary_large_image" />
-  <meta name="twitter:site" content="{{ twitter }}" />
+  <meta name="twitter:site" content="{{ url }}" />
```
jiwalker-usgs: `twitter:site` is supposed to be a twitter username `@user`

mwernimont:
looks simple enough

inst/templates/semantics.mustache
wdwatkins: wasn't sure on that

wdwatkins: Is there a template somewhere for what facebook should look like?

Issue #193, opened 2017-09-08
rebuild necessary items based on changes to viz.yaml non-file options


aappling-usgs:
first task: split the viz.yaml into pieces, write files behind the scenes that only include those pieces

second task: get dependencies hooked up so that items depend on their baby viz.yaml files

supports #103

ldecicco-USGS:
OK, for step 1, I'm thinking of doing the save in `getContentInfo`...does that sound correct?

aappling-usgs:
It needs to be called before the `make` targets that might depend on it; otherwise, they'll use old information. When & how would `getContentInfo` get called?

ldecicco-USGS:
Ohhhhh right, I think that is post-make

ldecicco-USGS:
Actually, if I run `createMakefiles` with `debug(getContentInfo)`, then I get pushed into the `getContentInfo` debugger, so I guess that is being called...

aappling-usgs:
Just remember that `getContentInfo` also gets called lots of other times, e.g., every time you run a fetch/process/visualize function (via `as.viz`). 

What about creating a new target in the main Makefile that depends on viz.yaml and is depended on by fetch, process, etc.?

ldecicco-USGS:
Talking myself through this....

I was thinking something like this:
```
getContentInfo <- function(viz.id, block=c('fetch','process','visualize','publish','resource'), no.match=c('stop','NA')) {
  if(length(viz.id) != 1) stop("viz.id must have length 1; see getContentInfos for alternatives")
  infos <- getContentInfos(viz.id, block=block)
  if(length(infos) == 0) {
    if(!missing(no.match) && is.na(no.match)) no.match <- 'NA'
    no.match <- match.arg(no.match)
    switch(
      no.match,
      'stop' = stop("no matches for viz.id '", paste0(viz.id, collapse="', '"), "' in block[s] ", paste0(block, collapse=", ")),
      'NA' = return(NA))
  }
  if(length(infos) > 1) stop("multiple matches for viz.id ", paste0(viz.id, collapse=", "), " in block[s] ", paste0(block, collapse=", "), ":\n", paste0("  ", names(infos), collapse="\n"))
  
  return.info <- checkConfig(info=infos[[1]], viz.id)
  
  return(infos[[1]])
}

checkConfig <- function(info,viz.id){
  file.config <- paste0(viz.id,".rds")
  full.config <- file.path("cache","config",file.config) 
[...]

PR #206, opened 2017-09-12
Rebuild

ldecicco-USGS:
Here's the start of #193

It adds a step to the Makefile, and creates a content.make file. I can run make("content"), and it seems to do what I expect. @aappling-usgs maybe you want to have a look at this while I try to start step "2" which is then having each task depend on the corresponding rds file that is generated in the `updateContent` function.


R/createMakefiles.R
```diff
@@ -61,6 +64,39 @@ createCleanRule <- function() {
...
+}
+
+#' Create call to the Content make file
+#'
+#' This includes an internal step "content" which is needed for triggering
+#' a rebuild if the viz.yaml file is changed.
+#'
+#' @keywords internal
+createContentMakefileCall <- function(){
+  paste('content: vizlab/make/content.make\n\tmake -f vizlab/make/content.make\nvizlab/make/content.make: viz.yaml ~/.vizlab/profile.yaml')
```
aappling-usgs: I like that this approach keeps the main Makefile simple, but it seems like a lot of extra text for a single additional workhorse target (`generate-content`). I don't anticipate many additional commands like this one, so I don't think we'd be risking an overly complex Makefile down the road...What would you think about skipping the step of creating the content.make file, and just putting this command right in the main Makefile?
```make
generate-content: viz.yaml
	${RSCRIPT}  -e 'library(vizlab); updateContent()'
```
(that command would take the place of 
```make
content: vizlab/make/content.make
	make -f vizlab/make/content.make
vizlab/make/content.make: viz.yaml ~/.vizlab/profile.yaml
```
)

aappling-usgs: Also, I really like the use of ${RSCRIPT} - way more readable.

aappling-usgs: Also, maybe a name other than `content` for the target? I think it's appropriate for the `updateContent()` function name but could be confusing to users focused on vizzie content (text, plots, etc.). How about going with `config` to be consistent with your new cache folder? maybe just `config`, or `parse-configs` or `snapshot-config`? Or taking a different tack, there's `parse-vizyaml`, `read-vizyaml`, or `parse-viz-config`...

R/createMakefiles.R
```diff
@@ -115,6 +146,81 @@ createBlockMakefile <- function(block=c('fetch','process','visualize','publish')
...
+#' 
+#' Creates 1 file per viz id with the content info. This is necessary to 
+#' update the project if only the viz.yaml is updated (otherwise make)
+#' won't trigger an update.
+#' 
+#' @param viz.id character[s] giving specific IDs of items to return, or missing
+#'   to return all content items
+#'   
+#' @export
+updateContentInfoFile <- function(viz.id){
```
aappling-usgs: this function looks just right

R/createMakefiles.R
```diff
@@ -115,6 +146,81 @@ createBlockMakefile <- function(block=c('fetch','process','visualize','publish')
...
+}
+
+#' updateContent
+#'
+#' Create rds files for each id's content
+#' @param block character name of the block in the viz.yaml for which to create a
+#'   makefile
+#'
+#' @export
+updateContent <- function(block=c('fetch','process','visualize','publish')){
```
aappling-usgs: :+1:

R/createMakefiles.R
```diff
@@ -61,6 +64,39 @@ createCleanRule <- function() {
...
 }
 
+#' Create the top shortcuts
+#'
+#' This includes an internal step "content" which is needed for triggering
+#' a rebuild if the viz.yaml file is changed.
+#'
+#' @keywords internal
+createShortcuts <- function(){
+  blocks <- c('content','fetch','process','visualize','publish')
```
aappling-usgs: i'm a bit uncomfortable with calling 'content' a block, even though this is a purely inward-facing function. one proposed solution is below: rather than creating and `make`ing the content.make file, maybe just put the `generate-content` recipe right in the main Makefile?

however, the work you've done here paves the way for adding an actual `globals` or `config` block to the viz.yaml, then writing a new globals.make/config.make file for processing those inputs. so if you like my idea for moving `generate-content`, maybe still leave this here and only take out `'content',` above? then when we get to globals, it'll just be a matter of dropping that in here.

R/createMakefiles.R
```diff
@@ -71,7 +70,7 @@ createCleanRule <- function() {
 #'
 #' @keywords internal
 createShortcuts <- function(){
-  blocks <- c('content','fetch','process','visualize','publish')
+  blocks <- c('config','fetch','process','visualize','publish')
```
aappling-usgs: so this now refers to the globals block to be created later, right?

ldecicco-USGS:
Getting the right target, depends combo (to add an additional depends) is trickier than I thought...

R/createMakefiles.R
aappling-usgs: i set myself up for confusion by suggesting 'config' for both globals and viz parameters... :confused:

aappling-usgs: oh! wait, i get it!

ldecicco-USGS: yeah, this is more behind the scene, "global" will be another section in the viz.yaml

aappling-usgs: cool. i was hung up on the variable name `blocks`, which i originally was using to describe those viz.yaml sections. but once i built the makefile with the new code it was clearer.

aappling-usgs:
This PR looks like a complete thought and a big step forward right now - you agree? Ready to merge?

ldecicco-USGS:
I think it's good, I haven't figured out a good way to test it other than fiddling with the example project. I'd say merge it especially since you are working on similar files, and I'll make an issue (and assign it to myself) to work out some basic tests

Issue #207, opened 2017-09-13
fetching data/xx interacts poorly with any additional dependencies


aappling-usgs:
There's a scenario where having all fetch items depend on their respective config files is a problem: when the fetch item is a file in data and therefore never gets copied into the cache, its timestamp never changes. So if it depends on cache/config/xx.rds, it gets 'fetched' every time. The fetch is just a few shell & function calls, no real action, but it still adds to every build time by a second or two. My example is iris-data in USGS-VIZLAB/example; it may be less common in real vizzies, but it didn't take long to find this real example: https://github.com/USGS-VIZLAB/water-use/blob/master/viz.yaml#L121.

It could be that we just need to remove the config/xx.rds dependency for items already in data/. or maybe we `touch` the data file after a change to the config, or `touch` the config to appear older than the data file? Or maybe we require & check that the config doesn't contain any parameters (because data can't use them) and _then_ remove the config/xx.rds dependency? Or maybe we copy data/ files into cache/ and don't permit targets in the data/ folder? I don't see a clear winner yet.

This issue emerged during https://github.com/USGS-VIZLAB/vizlab/issues/193 but could be addressed independently of it, depending on the solution we choose.

aappling-usgs:
possible solutions
* status file to represent data files
OR
* have fetch.file actually update the timestamp of data/iris.csv

laura points out that if we don't fix this this sprint, the effect is to allow builds to take longer than they would otherwise. could be worse.

jread-usgs:
not sure if I am not interpreting this correctly, but we also have a common fetch pattern that gets re-run each time when we use packaged geospatial data from the `maps` package. See irma or harvey's use of counties, states, and islands. I think these `Spatial` objects are re-built every time we run fetch even though the underlying package data used to create them does not change (i.e., the version of `maps` or `mapdata` packages). Not a big deal, but this does add a few seconds per target and may be bigger for other vizzies (maybe all US counties...)

aappling-usgs:
@jread-usgs - i think that particular fetch pattern can be explained by #192 and does not need to be re-addressed here, though specific confirmation that your use case has been resolved (in that issue) would be nice.

aappling-usgs:
I think this will not be a problem with remake. `mayfly_nymph` in the testviz is an example, because the data file doesn't get updated or copied into cache:
```yml
  -
    id: mayfly_nymph
    location: data/mayfly_nymph.csv
    fetcher: mayfly_nymph
    info: insect
[...]

PR #208, opened 2017-09-13
Improved config workflow

ldecicco-USGS:
This deals with #193 (1) and (3)


aappling-usgs:
nice, that really cleans it up!

aappling-usgs:
I believe this also sometimes addresses #193 (2), because now every call to `updateConfigInfoFile()` has the potential to update the .rds file and its timestamp.

Best case (and common) scenario:
You build iris-data once. You edit the configuration for iris-data (ignore https://github.com/USGS-VIZLAB/vizlab/issues/207 for now) in viz.yaml. You call `make -f vizlab/make/fetch.make iris-data`.  `make` notices that `cache/config/iris-data.rds` is out of date relative to `viz.yaml`, so it calls `updateConfigInfoFile`, whcih updates `cache/config/iris-data.rds`. Next time you build `iris-data`, `iris-data.rds` will already be up to date and won't result in any more calls to `updateConfigInfoFile` (until viz.yaml is updated again). Perfect!

Suboptimal (and also common) scenario:
You build iris-data once. You edit the configuration for other things, but not iris-data, in viz.yaml. You call `make -f vizlab/make/fetch.make iris-data`.  `make` notices that `cache/config/iris-data.rds` is out of date relative to `viz.yaml`, so it calls `updateConfigInfoFile`, and because iris-data config hasn't changed, `cache/config/iris-data.rds` is not altered. This is great because now `iris-data` won't rebuild, because it's already up to date relative to iris-data.rds. The suboptimal part is that every time you build iris-data without actually editing the iris-data info in viz.yaml, updateConfigInfoFile will get called again (and again will do nothing to iris-data.rds). So we're looking at a lot of calls to updateConfigInfoFile for this and other targets.

I think I'm OK will both the best-case and the suboptimal scenarios. There's probably some way to fix the latter, but I don't see an elegant one right now, and I think the costs to users will not be high. Let's roll with it, and we can consider a further optimization in the future if it turns out to be too slow in practice.

PR #212, opened 2017-09-13
Globals

ldecicco-USGS:
So now can have:
```
global:
  -
    id: plot-info
    height: 700
    width: 700
  -
    id: color-stuff
    points: green
```
then later:
```
visualize:
  -
    id: plot-cars
    location: figures/cars.png
    scripts: scripts/visualize/visualize.R
    visualizer: cars
    depends: [ "calc-cars", "cars-data","plot-info" ]
    mimetype: image/png
    title: "Speed vs. Braking Distance"
    alttext: "plot of ratio of speed to braking distance"
```
(see `plot-info` is in depends)

I'll write some tests tomorrow, if anyone wants to start reviewing before, have at it.


ldecicco-USGS:
This apparently now includes an `upstream master` pull I did that had Marty's footer work.

R/createMakefiles.R
```diff
@@ -363,6 +363,15 @@ createMakeItem.process <- function(item.info, ...) {
   createMakeRulePair(item.info, block='process', concat=TRUE)
 }
 
+#' \code{createMakeItem.global}: create makefile rules for an item in the
+#' global block of viz.yaml
+#'
+#' @rdname createMakeItem
+#' @export
+createMakeItem.global <- function(item.info, ...) {
```
aappling-usgs: This function creates rules like this in `global.make`:
```make
vizlab/make/config/plot-info.rds: 
	${RBATCH} "--args fun=global funargs=list(viz='plot-info')" \
	vizlab/make/callFunction.R vizlab/make/log/global/plot-info.Rout
```
where `global(viz='plot-info')` returns `invisible(NULL)`. Seems like a lot of infrastructure for something that's done by default for other sections - wouldn't the generic rule still apply, such that you could simply use the first line and no other rule specific to `plot-info`?:
```
plot-info: vizlab/make/config/plot-info.rds
```
relying on this generic rule:
```
vizlab/make/config/%.rds : viz.yaml
	${RSCRIPT}  -e "library(vizlab); updateConfigInfoFile('$*')"\
	> vizlab/make/log/config.Rout 2>&1
```
or does `global` do something different that I'm missing?

R/readData.R
```diff
@@ -96,6 +96,14 @@ readData.filepath <- function(viz){
   return(viz[['location']])
 }
 
+#' \code{readData.global} returns the file path or null if location is not set
+#'
+#' @rdname readData
+#' @export
+readData.global <- function(viz){
+  return(viz)
```
aappling-usgs: I see this for the example you provided:
```r
> readData('plot-info')
$id
[1] "plot-info"

$height
[1] 700

$width
[1] 700

$block
[1] "global"

attr(,"class")
[1] "global"
```

It's good that any information I want will be available. But this behavior provides some additional info (block, attr(,'class')) that vizlab users probably don't need and should't be seeing. What do you think about removing at least those two elements, possibly also the `id` because most `readData` items don't return the `id`? Also, what about taking out `location`, `scripts`, `reader`, `mimetype`, `depends`? I wonder if there's a list of the known keywords anywhere...

ldecicco-USGS: yeah, walker and I discussed removing the `block`. He removes it only in the publish viz stuff, so I didn't want to remove it because we weren't 100% sure there weren't other places where it might be needed. That being said, it would be pretty easy to take it out and see if anything breaks. I could take out `id` too, that makes sense. 

I'll test taking out the global class on that data...I think that's OK, but I did make a few "global" methods...I think the readData is the furthest downstream of anything so that should be fine.

What do you mean " taking out location, scripts, reader, mimetype, depends"? The struggle with this task was to make a step in the `vizlab` process that just passed on stuff from the yaml, so the user wouldn't need all that stuff (location, scripts, reader, mimetype, depends)....is it still hidden in there or something?

R/createMakefiles.R
ldecicco-USGS: So,does this look right?:
```make
# Rules

all: \
	plot-info \
	color-stuff \
	MakeMessages

vizlab/make/config/%.rds : viz.yaml
	${RSCRIPT}  -e "library(vizlab); updateConfigInfoFile('$*')"\
	> vizlab/make/log/config.Rout 2>&1
plot-info: vizlab/make/config/plot-info.rds
color-stuff: vizlab/make/config/color-stuff.rds

MakeMessages: 
	@echo ""
	@echo "global.make looks OK!"
	@echo ""

```

R/readData.R
aappling-usgs: Yeah, since readData is downstream of all the other generic functions, I think that's the right place to take out the global class.

When I suggested taking out keywords like location, scripts, or reader, I hadn't actually tried specifying a different reader, but I had been imagining that a global item could also refer to a small text or csv or yml file. Now that I'm trying it, I see that any choice of `reader` gets overridden and the item gets directed to `readData.global` every time. So your confusion makes sense - for globals, there currently are no keywords other than id and block. I guess that's not a bad way for globals to work...but is it the best way? Couldn't there be cases where someone creates a separate input file - text that gets used in multiple plot legends and the caption, say - and wants to call it global and read it as a yml? Or would that be too messy/hard to do?

ldecicco-USGS: In my head, that situation would then be a `fetch` process...whereas `global` is just a lightweight way to pass around stuff just in the yaml.  Since we totally have readers and infrastructure baked into `fetch`, I'm not sure what adding that same ability into `global` adds. 

Maybe we want to change the name `global` to something else?

aappling-usgs: Yeah, good point that `fetch` would be a fine place for such files. For renaming `globals`..., maybe `options`, `parameters`, or `specifications`?

R/createMakefiles.R
ldecicco-USGS: yup! giving it a shot now

aappling-usgs:
Nice work on 3eb0186!

R/readData.R
```diff
@@ -96,6 +96,16 @@ readData.filepath <- function(viz){
   return(viz[['location']])
 }
 
+#' \code{readData.parameter} returns the file path or null if location is not set
+#'
+#' @rdname readData
+#' @export
+readData.parameter <- function(viz){
+  viz <- viz[names(viz)[!(names(viz) %in% c("id","block"))]]
+  class(viz) <- 'list'
```
jiwalker-usgs: This should already be a list, I don't know that it should be classed in this way.

R/vizlab.R
```diff
@@ -23,7 +23,14 @@ as.viz.character <- function(x, ...) {
   if (length(viz) == 1 && is.na(viz)) {
     viz <- getResourceFromLibrary(x, no.match='stop')
   }
-  return(as.viz(viz))
+  viz <- as.viz(viz)
+  if("block" %in% names(viz)){
+    class(viz) <- c(class(viz), viz[["block"]])
```
jiwalker-usgs: The order of this should be flipped, the last class should be the most general (in this case `viz`) and the first should be the most specific.

R/vizlab.R
```diff
@@ -23,7 +23,14 @@ as.viz.character <- function(x, ...) {
...
+    if(viz[["block"]] == "parameter"){
+      class(viz) <- "parameter"
```
jiwalker-usgs: I think this should end up with class equal to `c("parameter", "viz")`, though I can't really think when we would need it to be a `viz` object.

tests/testthat/tests-readData.R
```diff
@@ -13,10 +13,14 @@ test_that("readData works as expected", {
 
   mayfly <<- readData(viz[['depends']][1])
   carData <<- readData(viz[['depends']][2])
+  plotData <<- readData(viz[['depends']][3])
```
jiwalker-usgs: We should probably use the newer `readDepends()` here at some point.

ldecicco-USGS:
image

aappling-usgs:
pretty!

ldecicco-USGS:
Full paths to footer thumbnails (also works for `publishLandingPage`), and the user can manually override. I personally think this PR is ready to merge

aappling-usgs:
coupla responses to jordan's comments first?

R/readData.R
ldecicco-USGS: 99.9% sure I had to put this in because it was coming in as a "parameter" class (that was added right by your next comment). Maybe if I flip it and it comes in as a "viz" object first, this won't be necessary. As to the 2nd comment...I'm pretty sure this is a pretty common way to class things...what are you suggesting? `as.list`?

jiwalker-usgs: I guess I didn't think about needing to remove the existing classes, this makes sense to me now.

tests/testthat/tests-readData.R
ldecicco-USGS: `readDepends` is tested below this.

Issue #161, opened 2017-02-28
Social media buttons


eread-usgs:

eread-usgs:
![screen shot 2017-02-28 at 3 56 01 pm](https://cloud.githubusercontent.com/assets/7794368/23435960/70f8f258-fdce-11e6-8ab0-dbe630217291.png)

eread-usgs:
something like this (above) from here: https://howmuch.net/articles/most-least-expensive-universities-in-usa

eread-usgs:
We want the buttons to stay at the top of the page while scrolling, and have them vertically stacked.

eread-usgs:
And, the embed button should show up here too.

jiwalker-usgs:
We will want to bring functionality from gages-through-ages to vizlab for use more commonly.

aappling-usgs:
should be able to opt in or opt out (e.g., google-analytics wouldn't want these buttons)

aappling-usgs:
TO DO:
- [ ] we need a publisher for social media that can pick up the required URLs for linking to facebook, twitter, and github.
- [ ] still working on placement for the icons from usability/prettiness perspectives, especially for mobile. do we want it just horizontal, or want ability to make it vertical on some of them? two different style sheets could give us a lot of flexibility here.

TABLED:
* reddit - no idea how this would work yet. tabling this one for now. where would it link? would each user who clicks on the icon be able to generate their own reddit post? ideally this would provide a link to an existing reddit on which you could post.

PR #213, opened 2017-09-14
simple social media bake in

mwernimont:


aappling-usgs:
does this close #161?

mwernimont:
Theres some discussion and some R work to be done still

aappling-usgs:
TO DO:
- [ ] we need a publisher for social media that can pick up the required URLs for linking to facebook, twitter, and github.
- [ ] still working on placement for the icons from usability/prettiness perspectives, especially for mobile. do we want it just horizontal, or want ability to make it vertical on some of them?

TABLED:
* reddit - no idea how this would work yet. tabling this one for now. where would it link? would each user who clicks on the icon be able to generate their own reddit post? ideally this would provide a link to an existing reddit on which you could post.

PR #216, opened 2017-09-15
adaptable footer

mwernimont:
Wanted the footer to adapt to what content it had in it. So if there are no blogs and 2 vizzies the vizzies take up the horizontal space instead of stacking on top of each other, at least until mobile.  

Tested a bunch of different combos and am pretty happy with the results I'm seeing.


jiwalker-usgs:
Can you change the target of this pull request to the 0.2.x branch? I missed this on your last PR, but we wanted to avoid messing up master with these larger changes.

mwernimont:
@jiwalker-usgs right branch?

jiwalker-usgs:
Yup, that is where we want it.

Issue #192, opened 2017-09-08
Refetch work


ldecicco-USGS:
Check the test, make sure it's working.

aappling-usgs:
this could be a bandaid on the way to fully implementing #102, or a complement to that future work.

specific example: we want a block like this to not run each time:
```  -
    id: precip-data
    fetcher: precip
    start.date: "2017-09-07"
    end.date: "2017-09-08"
    scripts: scripts/fetch/fetchPrecip.R
    location: cache/precip.csv
    mimetype: text/csv
    refetch: FALSE
    depends:
      counties: counties
    comment: disabling refetch to avoid too many GDP requests
```
and yet it seems to be running each time

aappling-usgs:
at first, refetch=TRUE was supposed to mean don't ever cache this
then later we thought refetch=FALSE was supposed to mean don't ever refetch this
clearly communication will need to be part of this solution

aappling-usgs:
here's what refetch claims to do right now (https://github.com/USGS-VIZLAB/vizlab/blob/0.2.x/vignettes/vizlabExample.Rmd#controlling-the-fetch-section):
> By default, the refetch option is set to TRUE in the viz.yaml for each section. vizlab checks the timestamps of online resources to see if the file or resource has changed since it was last fetched. If the timestamp has changed, or a timestamp cannot be obtained, the data will be fetched again. However, this may be undesireable in some scenarios. To force vizlab to use the cached versions of remote files, set refetch to FALSE in the relevant sections of the viz.yaml. Conversely, to force vizlab to refetch the data regardless of timestamps, set refetch to TRUE and delete the timestamps for each section. These are stored in vizlab/make/timestamps.

aappling-usgs:
experience in irma (quotes & summaries from slack)

1. how does one get a block like this to not run each time (because it still is):
```
  -
    id: precip-data
    fetcher: precip
    start.date: "2017-09-07"
    end.date: "2017-09-08"
    scripts: scripts/fetch/fetchPrecip.R
    location: cache/precip.csv
    mimetype: text/csv
    refetch: FALSE
    depends:
      counties: counties
    comment: disabling refetch to avoid too many GDP requests
```

2. "it was always a bit goofy I think", "I don't think it has ever worked as expected"

3. people forget that refetch=TRUE (e.g., they had it at FALSE but then pull as TRUE, don't notice) and then get frustrated that it's refetching

[...]

Issue #102, opened 2016-10-24
Concept: add ttl to fetchers


jiwalker-usgs:

jiwalker-usgs:
To alleviate some of the issues with fetchers refetching too often, add a time-to-live added to the fetch, such that it isn't a binary decision (fetch or not), but can refetch after an allotted amount of time.

jiwalker-usgs:
related to #96

aappling-usgs:
this code works. better copy it before i mess up the syntax.
```make
iris-data_timestamp: 
ifeq ($(shell echo $(shell Rscript -e "vizlab::exceededTimeToLive('iris-data')" 2> null)),TRUE)
	@echo "iris-data_timestamp: fetchTimestamp(viz='iris-data')"
	@export R_LIBS_USER=$(RLIBSUSER);\
	${RBATCH} "--args fun=fetchTimestamp funargs=list(viz='iris-data')" \
	vizlab/make/callFunction.R vizlab/make/log/fetch/iris-data_timestamp.Rout
else
	@echo "iris-data_timestamp: exceededTimeToLive('iris-data')=FALSE"
endif
```

aappling-usgs:
and yet it still seems that the file is building no matter whether we fetch the timestamp or not?
```
$ make -f vizlab/make/fetch.make iris-data
iris-data_timestamp: fetchTimestamp(viz='iris-data')
data/iris.csv: fetch(viz='iris-data')

$ make -f vizlab/make/fetch.make iris-data
iris-data_timestamp: exceededTimeToLive('iris-data')=FALSE
data/iris.csv: fetch(viz='iris-data')
```

aappling-usgs:
iris-data just isn't going to work until resolution of #207, but i think cars-data works on my fork right now...
```yaml
timetolive:
  cars-data: 20 secs
```

```make
$ rm vizlab/make/timestamps/cars-data cache/cars.csv

$ ls -l vizlab/make/timestamps/cars-data cache/cars.csv
ls: cannot access 'vizlab/make/timestamps/cars-data': No such file or directory
ls: cannot access 'cache/cars.csv': No such file or directory

$ make -f vizlab/make/fetch.make cars-data
cars-data_timestamp: fetchTimestamp(viz='cars-data')
cache/cars.csv: fetch(viz='cars-data')

$ ls -l vizlab/make/timestamps/cars-data cache/cars.csv
[...]

PR #217, opened 2017-09-15
require fetchTimestamp but with timetolive opt-out

aappling-usgs:
Addresses issues #192 and #102, which are closely related

* require absence of `refetch` and `fetchTimestamp` in viz.yaml fetch item chunks
* require presence of `fetchTimestamp.xx` method for every fetcher `xx`
* built-in `fetchTimestamp.file` method writes a timestamp file at least once
* slightly easier-to-use `writeTimestamp` function for custom `fetchTimestamp` methods
* implemented a timetolive option for every fetch file in a new preferences.yaml file that users can put in their home project directory (and should not git commit, because this file is for _local_ preferences)
* scenarios should work as in the table at https://github.com/USGS-VIZLAB/vizlab/issues/192#issuecomment-329681645


R/createMakefiles.R
```diff
@@ -329,23 +334,43 @@ createMakeItem.fetch <- function(item.info, ...) {
...
+    phony.timestamp <- createMakeBatchRule(
       target=timestamp.id,
       fun='fetchTimestamp',
       funargs=c(viz=squote(item.info$id)),
       scripts=item.info$scripts,
       logfile=paste0('fetch/', timestamp.id, '.Rout'))
+    # interleave a gnu make conditional (I hope it works on all systems!!) for exceedance of timetolive
+    phony.timestamp.split <- strsplit(phony.timestamp, split='\n')[[1]]
+    phony.timestamp.full <- c(phony.timestamp.split[1],
+      sprintf("ifeq ($(shell echo $(shell Rscript -e \"vizlab::exceededTimeToLive('%s')\" 2> null)),TRUE)", item.info$id),
```
aappling-usgs: whoops, just realized there's now a file named `null` in my repo, thought I was sending to empty. i probably meant `/dev/null`

ldecicco-USGS: while you are fixing that, don't forget to add `importFrom("utils", "getS3method")` or however you want to do that to get Travis happy

aappling-usgs: :+1:

ldecicco-USGS:
FYI, it *looks* pretty awesome, if you are happy with it, I'd vote merging and if we discover issues make new issues. We'll want to merge this and https://github.com/USGS-VIZLAB/example/pull/18

PR #218, opened 2017-09-15
Svg inject

jiwalker-usgs:
One issue remaining is that transitive dependencies don't resolve, so injection requires adding jquery, svg-injector, and vizlab-js to the page resources.  There is an example using this in example.


ldecicco-USGS:
"There is an example using this in example." You mean the `example` viz? Could you push that up (or did I miss it...)?

jiwalker-usgs:
It actually added to the previously open PR https://github.com/USGS-VIZLAB/example/pull/12

PR #227, opened 2017-09-19
resolves #226 for common use cases

### LINKED ISSUE #226: quick way to open log files
would be nice if you could just type `getLog('hurricane-track')` or `getLog('hurricane-track_timestamp')` and have the log file open in your editor or console

aappling-usgs:
jordan, will this functionality work for you? there might be bugs for uncommon use cases i haven't thought of, but i think this will be enough to help us through maria.

* Attempts to find a viz .Rout file based on a character id.
* Prints a message with the tail of the log file if `tail=TRUE`.
* Opens the file in an editor if `open=TRUE`.
* Always returns the entire log file as a character vector.


jread-usgs:
I think so - I am assuming this works even if I am running make in terminal, and then hopping into RStudio to run this. I often use terminal because I can easily toggle which stages I build (e.g., `make publish` vs `make`) and I don't know a quick way to do that within RStudio

aappling-usgs:
yes, that's why i use the terminal too, and this should work fine in R even if you're bouncing. It's just looking at files.

PR #230, opened 2017-09-20
rework timestamp fetching

aappling-usgs:
* create & export two fetchTimestamp functions meant to be assigned to custom fetchTimestamp.fetcher methods (`alwaysCurrent` and `neverCurrent`)
* make timestamp-managing helper functions friendlier, better documented
* write new helper function, `urlFetcher`, to wrap up some of the dummy fetcher creation that was being done in hurricane-irma, -maria


jiwalker-usgs:
It seems the travis test is complaining about the sciencebase timestamp, is this something to worry about?

aappling-usgs:
the test passes locally...it's probably test design rather than a package problem...but i can spend a few mins trying to fix it after this sprint check-in

aappling-usgs:
will debug after this

PR #231, opened 2017-09-21
inject on orientation change

jiwalker-usgs:


ldecicco-USGS:
Silly sciencebase not knowing time zones...

PR #232, opened 2017-09-21
viz.yaml chunk dependency + sciencebase timezone fix

aappling-usgs:
* updateConfigInfoFile now updates when values change (issue from slack/maria: need _any_ change to `mobile-metadata` to cause rebuild of `view-limits-mobile`)
* SB dateUploadeds are UTC - hopefully will fix travis build?

### DIFF
--- a/R/createMakefiles.R
+++ b/R/createMakefiles.R
@@ -186,7 +186,7 @@ updateConfigInfoFile <- function(viz.id){
   full.config <- file.path("vizlab/make/config",file.config) 
   if(file.exists(full.config)){
     orig.info <- readRDS(full.config)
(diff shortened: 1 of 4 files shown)


R/createMakefiles.R
```diff
@@ -186,7 +186,7 @@ updateConfigInfoFile <- function(viz.id){
   full.config <- file.path("vizlab/make/config",file.config) 
   if(file.exists(full.config)){
     orig.info <- readRDS(full.config)
-    if(!all(orig.info %in% info)){
+    if(!isTRUE(all.equal(orig.info, info))) {
```
ldecicco-USGS: :+1:

PR #236, opened 2017-09-25
Wrapping up markdown

### LINKED ISSUE #214: close off markdown example
https://github.com/USGS-VIZLAB/example/pull/12

jiwalker-usgs:
fixes #214 

Also note that attempts to put markdown inline, straight in the viz.yaml were made and not supported.  This use case may be something for the future.

### DIFF
--- a/tests/testthat/tests-markdown.R
+++ b/tests/testthat/tests-markdown.R
@@ -19,4 +19,20 @@ test_that("markdown in context becomes proper html", {
   expect_match(fragment$list, "<li>or numbered</li>")
 
 })
+
+test_that("markdown in viz converts to html", {
+  viz <- as.viz(list(
+    id = "md_inline",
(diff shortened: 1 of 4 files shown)


tests/testthat/tests-markdown.R
```diff
@@ -19,4 +19,20 @@ test_that("markdown in context becomes proper html", {
   expect_match(fragment$list, "<li>or numbered</li>")
 
 })
+
+test_that("markdown in viz converts to html", {
+  viz <- as.viz(list(
+    id = "md_inline",
```
aappling-usgs: why is this named "inline"? i thought you said inline isn't supported yet?

jiwalker-usgs: You're correct, I started off trying to test inline and it didn't work, I never changed the ID.  Will do that now.

PR #237, opened 2017-09-28
readData.json

lindsaycarr:
Used pretty frequently in these hurricane vizzies so makes sense to include in the package. Wasn't sure if mimetype needed to be included. I believe it's `mimetype/javascript`?


R/readData.R
```diff
@@ -189,6 +189,19 @@ readData.inline <- function(viz) {
...
+#' \code{readData.json} reads json files
+#'
+#' @rdname readData
+#' @return single value or list of values
+#' @export
+readData.json <- function(viz) {
+  required <- c("data")
+  checkRequired(viz, required)
+  
+  data <- jsonlite::fromJSON(viz[['location']])
```
aappling-usgs: because `jsonlite` is in Suggests rather than Imports, i think we want a check above this line to tell the user if jsonlite isn't installed - something like:
```r
has_jsonlite <- suppressMessages(suppressWarnings(require(jsonlite)))
if(!has_jsonlite) { 
  stop("jsonlite package is required to read JSON data (mimetype='xxxx', reader='json')")
}
```
(I used to use just `require(jsonlite, quietly=TRUE)` for checks like the above, but just realized that it's not silent despite the behavior implied by the documentation for `quietly`. so suggesting this longer message/warning suppression as given in https://stackoverflow.com/questions/14834841/when-does-quietly-true-actually-work-in-the-require-function)

R/readData.R
```diff
@@ -189,6 +189,19 @@ readData.inline <- function(viz) {
   return(data)
 }
 
+#' \code{readData.json} reads json files
+#'
+#' @rdname readData
+#' @return single value or list of values
```
aappling-usgs: this return value becomes confusing in the merged `readData` documentation (see below - i'll call it out). here your choices are to label this particular `@return` with `readData.json: ....` or `json: ...`, or simply to exclude the return value description in this method because it's already defined for the generic method (i'd recommend the latter option, because readData.json doesn't return anything especially different from other readData methods).

man/readData.Rd
```diff
@@ -41,12 +42,16 @@ readData(viz)
...
+\method{readData}{json}(viz)
 }
 \arguments{
 \item{viz}{vizlab object or identifier}
 }
 \value{
 single value or list of values
+
+single value or list of values
 }
```
aappling-usgs: here's where the multiple `@return` tags in the same `@rdname readData` help file get confusing: see how they're duplicated:
```
 \value{
  single value or list of values

 single value or list of values
  }
```

lindsaycarr:
I guess I'm not quite sure what exactly `mimetype` is doing here. In some `readData` functions it is a control for how to read the type of data, but for json, there's only the one way.

R/readData.R
```diff
@@ -201,7 +201,7 @@ readData.json <- function(viz) {
     stop("jsonlite package is required to read JSON data (mimetype='application/javascript', reader='json')")
   }
   
-  required <- c("data")
+  required <- c("location")
```
aappling-usgs: oops! good catch

aappling-usgs:
I think `mimetype` would mostly be useful here: https://github.com/USGS-VIZLAB/vizlab/blob/master/R/readData.R#L200-L216 to help with routing the default `readData` - adding it to that lookup would allow vizlab users to specify mimetype only and have the reader get discovered by this function.

aappling-usgs:
so would get added here: https://github.com/USGS-VIZLAB/vizlab/blob/67e4a89cb839c9663d8533de996de4e757c5eb76/inst/mimetypes.default.yaml to match the json reader to the mimetype.

i'm seeing `application/json` for json mimetype (https://stackoverflow.com/a/477819/3203184)

lindsaycarr:
hmmm `application/javascript` is what I saw in hurricane-irma: https://github.com/USGS-VIZLAB/hurricane-irma/blob/master/viz.yaml#L353-L361

ldecicco-USGS:
That was a hack (at least it was in Harvey) because we didn't have a proper json reader!

lindsaycarr:
ah OK, so it just propagated through to maria

ldecicco-USGS:
https://github.com/USGS-VIZLAB/hurricane-matthew/blob/master/viz.yaml#L219

Issue #261, opened 2017-10-03
add USGS & Google tracking code to vizlab landing page


aappling-usgs:
the vizlab landing page tracking ID is currently mixed up with the training pages tracking ID. that's no good. also see #137, where we'll be doing some of the same for the vizzy pages

PR #268, opened 2017-10-04
add google analytics to landing

wdwatkins:


aappling-usgs:
This would resolve https://github.com/USGS-VIZLAB/vizlab/issues/261, right?

wdwatkins:
yes

wdwatkins:
Looking at this again, I'm not sure what this index.html is actually doing.  It isn't a template, and it isn't what gets rendered.

aappling-usgs:
Yikes! Good to know. Glad to see you and Marty making progress on this at https://github.com/USGS-VIZLAB/vizlab/pull/269.

ldecicco-USGS:
When you run the function, you have to look in the landing folder, then the target folder...then index.html. For whatever reason, there's some *other* index.html in the landing folder just to throw us off I think.

wdwatkins:
Yeah, I'm deleting the decoy index.html in my current PR

Issue #137, opened 2016-12-19
add USGS and update Google analytics scripts


wdwatkins:

ldecicco-USGS:
https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/templates/fullPage.mustache#L31

should be:
`
ga('create', '{{ . }}', 'usgs.gov');
```
I think?

**Comment by ldecicco-USGS** [comment]
And just need to add:
```
<script type="application/javascript" src="https://www2.usgs.gov/scripts/analytics/usgs-analytics.js"></script>
```
somewhere in there too

aappling-usgs:
@wdwatkins - is this issue obsolete or still a good idea?

ldecicco-USGS:
Be sure to check with Jim too, at one point we were missing the USGS and DOI analytics. I'm not sure they are mandatory, but when things go viral, they would be noticed even more if we had the right scripts set up.

wdwatkins:
I'm not sure on Laura's first comment, but that script seems to indicate it should be used.  https://www2.usgs.gov/scripts/analytics/usgs-analytics.js

aappling-usgs:
jim says, "yeah, that's actually essential, required by usgs, doi, and omb policy" - we will prioritize this.

wdwatkins:
On semi-related topic, there is a new implementation of Google's analytics script that we might want to start using for our own analytics.  It is now the default option that google tells you to use when you make a new analytics ID.

aappling-usgs:
Keeping up sounds good. Are we talking a 1-2 line edit for that, too? Do you have a link you could share here?

wdwatkins:
https://developers.google.com/analytics/devguides/collection/gtagjs/migration

aappling-usgs:
there are also analytics buried within the javascript functions attached to the VIZLAB javascript namespace - those will need updating, too

jread-usgs:
the things I was talking about include `vizlab` scoped GA functions. So, from `maria`:
![image](https://user-images.githubusercontent.com/2349007/31462974-15f13bc8-ae94-11e7-81af-0f7c8fc259f5.png)

jread-usgs:
those functions are in `/inst/js/vizlab.js`

ldecicco-USGS:
- [ ] usgs-analytics.js is already in the mustache
- [ ] the line `ga('create', '{{ . }}', 'auto');` is in our script, but seems to work?
[...]

PR #269, opened 2017-10-06
landing analytics v2

wdwatkins:


mwernimont:
Hmm...you don't usually put a script tag inside of a header, you either have them in the head of the page, or just before the body tag finishes.  I don't think this is where we would want this script tag.

wdwatkins:
@mwernimont in the vizzies or tags mustache then?

wdwatkins:
Doesn't seem like there is a general page template in the `inst/landing` stuff.  Maybe it is using the default vizlab template?

mwernimont:
In inst/templates/fullPage.mustache looks to be the best bet, theres already a GA analytics script in there too.

wdwatkins:
Ok, yeah we should probably update the existing GA code and just make sure the landing page function utilizes that part of the template.

wdwatkins:
Ok @mwernimont, should be better now

wdwatkins:
This will add the USGS analytics to all pages, so partially fulfills #137.  I will leave the rest of that issue to a future PR, which involve modifying all the templates that have GA baked in.

R/publish.R
```diff
@@ -322,7 +322,8 @@ publish.landing <- function(viz){
   pageviz$context <- list(sections = c("owiNav", "header", names(viz_info)), #names of section ids
                           resources = c("landingCSS", "owiCSS", "jquery", "appJS"),
                           header = "usgsHeader",
-                          footer = "usgsFooter")
+                          footer = "usgsFooter", 
+                          info = list(`analytics-id` = "UA-78530187-11"))
```
wdwatkins: Not sure if we want to hard code this in or not?  In theory we wouldn't ever change this, but it could be passed in.

Issue #148, opened 2017-02-21
dynamic hover


eread-usgs:

eread-usgs:
Re-using some SVG JS stuff- make this a part of the package that is re-usable so that we dont have to C/P each time. Also improve function- get Mary on this? Needs more in order to be a robust hover.  Needs cleanup from JS person.

eread-usgs:
HI @mbucknell I think this is the task associated with the work thought you'd bring a lot to- just adding you as an assignee for tracking. This issue still needs more detail, which you can hopefully get from @jiwalker-usgs.

aappling-usgs:
This issue was also referenced as "good JS function for tooltips that we can reuse" in #98

jread-usgs:
I thought @mbucknell did this work. Did it not end up in the package? Also, the interns refined it in their summer work and we could likely borrow from their solution.

aappling-usgs:
Oh, great! I don't know where to look for it - could you point out some of the relevant code?

aappling-usgs:
Found it: https://github.com/USGS-VIZLAB/vizlab/pull/168

jread-usgs:
and https://github.com/USGS-VIZLAB/active-flood-viz/blob/b70fc869fc08f66e3f7506e0803ad3157744c966/floodviz/static/js/tooltip.js @aappling-usgs

aappling-usgs:
Next step on this issue is to look at the interns' work at https://github.com/USGS-VIZLAB/active-flood-viz/blob/master/floodviz/static/js/tooltip.js and https://github.com/USGS-VIZLAB/active-flood-viz/blob/master/floodviz/static/js/map.js, then import innovations from there into https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/js/svg.js

jread-usgs:
Overview of interns' work: 

1. added awareness of the "top" of the svg to the tooltip (was only adjusted for sliding off right and left before)
2. added D3 to the tooltip code which makes the select and modification of the tooltip cleaner
3. parameterized the triangle of the tooltip, so that is adjustable
4. cleaner code and better documentation in general

aappling-usgs:
Mary also mentioned that the current hover functionality wasn't working as of Irma - there may be some bugs or incompatibilities that could be fixed

aappling-usgs:
jake could be a PR reviewer - he's looking for interactivity skills

aappling-usgs:
[...]

PR #272, opened 2017-10-12
Tooltip code

jread-usgs:
This is a start on #148 
 - we now bring in d3 into vizlab minified, and can access it as a resource
 - the tooltip code we have been copypasta is now in the package, but updated for d3
 - instead of the tooltip becoming invisible on mouseover, it is just destroyed and rebuilt when needed
 - the tooltip components (tip, text, box) no longer need to be created in a `visualize` step, they are instead created in this function, assuming `#tooltip-group` exists (errors if not)
 
I haven't added any of the improvements Mary made to the code in #168 or the improvements the interns made this summer as noted in #148. But this is now functional and included in the package. To use it, update your viz.yaml:
image


inst/js/tooltip.js
```diff
@@ -0,0 +1,45 @@
...
+      clearTimeout(hoverTimer); //stop when off area
+    }
+  } else {
+    var tooltip_bg = tipG.append('rect').attr("id", "tooltip-box").attr("height", 24).attr("class", "tooltip-box");
+    var tool_pt = tipG.append('path').attr("id", "tooltip-point").attr("d", "M-6,-11 l6,10 l6,-11").attr("class","tooltip-box");
+    var tooltip = tipG.append('text').attr("id", "tooltip-text").attr("dy","-1.1em").attr('text-anchor',"middle").attr("class","tooltip-text-label svg-text").text(text);
+    pt = cursorPoint(evt);
+    pt.x = Math.round(pt.x);
+    pt.y = Math.round(pt.y);
+    var svgWidth = Number(svg.getAttribute("viewBox").split(" ")[2]);
```
aappling-usgs: does this line require that `svg` is in the global environment? and if so, is the existence of `svg` guaranteed by vizlab? what happens if a viz has multiple svg figures?

inst/js/tooltip.js
```diff
@@ -0,0 +1,45 @@
+var hoverTimer = null;
+var hoverDelay = 400; //ms
+  
+function hovertext(text, evt){
+  var tipG = d3.selectAll('#tooltip-group');
```
aappling-usgs: this makes sense, but just curious - is there a jquery alternative? seems like there are tons of selectors: https://api.jquery.com/category/selectors/

jread-usgs: yes, we could make this selection w/ jquery too

inst/js/tooltip.js
jread-usgs: Good question - that happens in animate.js right now, but I don't know how we should handle multiple svgs within one page, which seems likely in the future...

aappling-usgs: is it an option to pass in the svg as an argument to `function hovertext(text, evt){`? unless it's super easy, i think we can just create another issue for this and keep chugging, but agreed that it'd be nice to address at some point.

aappling-usgs: created an issue at https://github.com/USGS-VIZLAB/vizlab/issues/273

PR #274, opened 2017-10-12
Retina Favicon.

mwernimont:
Was able to create a favicon that will switch based on what size is needed. Found you can make a single favicon file with several sizes from https://www.smooster.com/en/blog/how-to-create-a-favicon-for-retina.html using http://www.xiconeditor.com/


jread-usgs:
cool: "Favicon is a word coined from the words “favorite” and “icon”." never knew. 

Looks good. it wasn't clear from the help if we actually need more than one ico in the viz. Do we? or does one work and the non-retina views deal with the higher-res one appropriately? 

I see one new image in the file, so is nothing needed (logic) to switch between multiple icos if they are needed?

mwernimont:
The one file contains 4 different ico files at different sizes.  The browser will talk to the ico and and the ico will feed the appropriate size to the browser.  At least that is what I'm reading.  So this should cover any situation we have I believe @jread-usgs

jread-usgs:
oh, ok. Limitation of GH's "view" just showed me one.

mwernimont:
Well it is one file. Idk how it all works, but it one file with multiple layer, and it decides which layer shows up.  Kinda new to me, so if it does work it should be sweet.  Way better than 4 different files floating around.

Issue #253, opened 2017-10-03
favicon for landing page


aappling-usgs:
![image](https://user-images.githubusercontent.com/12039957/31137371-eedf11ec-a81f-11e7-9cd7-9266830ec1d9.png)

pulling this task out of #124

mwernimont:
Solved by #275

PR #275, opened 2017-10-12
#253 favicon for landing page

mwernimont:
Added the vizlab favicon to the landing page publisher and to the landing page viz.yaml.  The favicon is in the vizlab resources yaml, so was pretty easy to call out to.


mwernimont:
Man, weren't we at 77% at one point?

Issue #205, opened 2017-09-12
use dssecrets for authentication


aappling-usgs:
this should be possible for both S3 and ScienceBase

aappling-usgs:
after completing this issue, should be able to delete `storeSBcreds`

aappling-usgs:
i think this issue is complete except for deleting/revising this docs line: https://github.com/USGS-VIZLAB/vizlab/blob/c06eb97ec1b670359236ed574ac4441ae23d5ecf/R/authRemote.R#L7

PR #276, opened 2017-10-13
switch from storeSBcreds to dssecrets, secret

### LINKED ISSUE #125: storeSBcreds should keep password hidden
Change to the way authenticate:

https://github.com/USGS-R/sbtools/blob/master/R/authenticate_sb.R#L59

does log in, so we don't have to see our passwords show up on the screen.

aappling-usgs:
Delete `storeSBcreds` function and associated ScienceBase login method. In its place, try a cascade of login options from already-logged-in (this will be possible once we switch to remake) to `dssecrets` to a personal `secret` vault to a warning without authentication.

resolves #125 and sciencebase side of #205


aappling-usgs:
This was a low-priority item but rose in priority when they removed public items from ScienceBase. Currently not even USGS-VIZLAB/example works without credentials matching CuyahogaTDS.csv. Rather than get Jake and Sam started on storeSBcreds, I figured we should go for a more secure solution.

PR #277, opened 2017-10-13
tweak console error message if jquery not found

dblodgett-usgs:


jread-usgs:
looks good to me :+1:

PR #282, opened 2017-10-14
Tooltip

### LINKED ISSUE #273: let hoverText handle multiple svgs on a page
See discussion at https://github.com/USGS-VIZLAB/vizlab/pull/272#discussion_r144186153 - right now tooltip.js won't work with multiple svgs, which is fine for some vizzies but could eventually bite us.

jread-usgs:
Fixes #273 
Now will PR on `example` viz to add this functionality

### DIFF
--- a/inst/js/tooltip.js
+++ b/inst/js/tooltip.js
@@ -22,27 +22,34 @@ function hovertext(text, evt){
     var svgWidth = Number(thisSVG.getAttribute("viewBox").split(" ")[2]);
     var textLength;
     var halfLength;
+    var textHeight;
+    var textBuffer;
     var tooltipX;
-    var tooltip_bg = tipG.append('rect').attr("id", "tooltip-box").attr("height", 24).attr("class", "tooltip-box");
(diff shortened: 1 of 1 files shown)


inst/js/tooltip.js
```diff
@@ -1,44 +1,55 @@
...
-  var tipG = d3.selectAll('#tooltip-group');
-  if (tipG.empty()) {
-    console.error("a #tooltip-group <g/> element is required");
-  }  
-  // remove all children of the tooltip group:
-  tipG.selectAll("*").remove();
+  
+  // remove all children from any tooltip group:
+  d3.selectAll("#tooltip-group").selectAll("*").remove();
+  // does the above work to clear out multiple tooltips from multiple svgs?
```
aappling-usgs: i think it does - i commented out line 7 in developer tools, created a bunch of tooltips on the 2 svgs, and then called this line before and after calling `d3.selectAll("#tooltip-group").selectAll("*")`. all of the tooltips in both groups disappeared after the `remove()` call. are you confident now, too, and if so how about removing this line?

jread-usgs: yes, confident - how about I remove it in #284 ?

aappling-usgs: :+1:

PR #284, opened 2017-10-15
Tooltip path

jread-usgs:
This is an add-on to #282 , but makes the tooltip box as a single <path> instead of a <rect> (for the box) and a <path> for the triangle tip. 

This gets rid of the little bit of separation or overlap between the triangle and the box that appears in other vizzies: 
image

### DIFF
--- a/inst/js/tooltip.js
+++ b/inst/js/tooltip.js
@@ -22,27 +22,34 @@ function hovertext(text, evt){
     var svgWidth = Number(thisSVG.getAttribute("viewBox").split(" ")[2]);
     var textLength;
     var halfLength;
+    var textHeight;
+    var textBuffer;
     var tooltipX;
-    var tooltip_bg = tipG.append('rect').attr("id", "tooltip-box").attr("height", 24).attr("class", "tooltip-box");
-    var tool_pt = tipG.append('path').attr("id", "tooltip-point").attr("d", "M-6,-11 l6,10 l6,-11").attr("class","tooltip-box");
-    var tooltip = tipG.append('text').attr("id", "tooltip-text").attr("dy","-1.1em").attr('text-anchor',"middle").attr("class","tooltip-text-label svg-text").text(text);
+    var tooltip_bg = tipG.append('rect').attr("id", "tooltip-box").attr("class", "tooltip-box");
+    var tool_pt = tipG.append('path').attr("id", "tooltip-point").attr("d", "M-6,-12 l6,10.5 l6,-10.5").attr("class","tooltip-box");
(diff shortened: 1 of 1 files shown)


dblodgett-usgs:
👍 I pulled down your branch on top of the latest here and it works great!

screen shot 2017-10-15 at 10 26 59 am

inst/js/tooltip.js
```diff
@@ -1,44 +1,69 @@
...
+    var textBuffer;
     var tooltipX;
-    var tooltip_bg = tipG.append('rect').attr("id", "tooltip-box").attr("height", 24).attr("class", "tooltip-box");
-    var tool_pt = tipG.append('path').attr("id", "tooltip-point").attr("d", "M-6,-11 l6,10 l6,-11").attr("class","tooltip-box");
-    var tooltip = tipG.append('text').attr("id", "tooltip-text").attr("dy","-1.1em").attr('text-anchor',"middle").attr("class","tooltip-text-label svg-text").text(text);
+    var tooltip_bg;
+    var tooltip;
+    tooltip_bg= tipG.append('path').
+      attr("id", "tooltip-box").
+      attr("class", "tooltip-box");
```
aappling-usgs: good idea to get this on multiple lines.

the convention i've been seeing is to put the dot at the beginning of each new line (e.g. https://d3js.org/#introduction) - do you have an opinion on where the dots should go, and why?

also, it looks to me like there's nothing stopping you from continuing to declare and define each variable in a single command, which seems a little cleaner to me. Like this:
```
var tooltip_bg = tipG.append('path')
    .attr("id", "tooltip-box")
    .attr("class", "tooltip-box");
```
what do you think about doing it that way instead? are there reasons not to?

inst/js/tooltip.js
```diff
@@ -1,44 +1,69 @@
...
+      text(text);
     
-    textLength = Math.round(tooltip.node().getComputedTextLength());
+    textBox = tooltip.node().getBBox();
+    textLength = Math.round(textBox.width);
+    textHeight = Math.round(textBox.height);
     halfLength = textLength / 2;
+    textBuffer = 6;
+    var tipYoffset = -2; // so that the tip is slightly above the mouse location
+    var tipTriangle = {x:6, y:10};
```
aappling-usgs: good idea to extract & name all these constants

inst/js/tooltip.js
jread-usgs: I was following Mary's style for the `var varname;` style ones. But the PR is a mix of both. Not sure why to do it in two lines?

jread-usgs: I also prefer the .attr() line starts. I will clean that up.

PR #286, opened 2017-10-15
Adding readData.shp and readData.svg

### LINKED ISSUE #240: read.shp
already implemented at https://github.com/USGS-VIZLAB/hurricane-maria/blob/master/scripts/read/read.shp.R, just need to check it over and add it to vizlab once it's sufficiently general for our needs

### LINKED ISSUE #241: read.svg
already implemented at https://github.com/USGS-VIZLAB/hurricane-maria/blob/master/scripts/read/read.svg.R, just need to read it over and add it in if it's sufficiently general for our needs

jzwart:
Adding reads for shapefiles and svg to vizlab, modified from Hurricane Maria viz. Fixes #240 and fixes #241 and is more generic than readers for Maria viz. More specific shapefile readers to be made for [stormviz ](https://github.com/USGS-VIZLAB/vizstorm/issues/19). 

I wasn't sure if svg reader should be left named as readData.svg or readData.svg_map. I thought the map was too specific but maybe that's all we use svg readers for, I just wasn't sure.


R/readData.R
```diff
@@ -207,6 +207,50 @@ readData.json <- function(viz) {
...
+#' 
+#' assumes only one .shp file in zip folder
+#' 
+#' @rdname readData
+#' @export
+readData.shp <- function(viz){
+  required <- c('location')
+  checkRequired(viz, required)
+  
+  has_rgdal <- suppressMessages(suppressWarnings(require(rgdal)))
```
aappling-usgs: i just ran across a similar error to what you're getting here in the travis build:
```
'library' or 'require' calls in package code:
  ‘jsonlite’ ‘rgdal’ ‘xml2’
  Please use :: or requireNamespace() instead.
```

i fixed it by switching from `require` to `requireNamespace` (as at https://github.com/USGS-VIZLAB/vizlab/pull/276/commits/82ae5df7e32925a77fbc4e7f5c331b87f75318c2#diff-338de3777eea962e9cbbb81c4639e7e5R49):
```
has_rgdal <- requireNamespace('rgdal', quietly=TRUE)
```

aappling-usgs: and to get travis building cleanly you'll also probably need to add rgdal to the list of suggested packages in the DESCRIPTION (https://github.com/USGS-VIZLAB/vizlab/blob/master/DESCRIPTION#L32-L40)

R/readData.R
```diff
@@ -207,6 +207,50 @@ readData.json <- function(viz) {
...
+
+#' \code{readData.svg_map} reads svg
+#' 
+#' @rdname readData
+#' @export
+readData.svg <- function(viz){
+  required <- c('location') 
+  checkRequired(viz, required)
+  
+  has_xml2 <- suppressMessages(suppressWarnings(require(xml2)))
```
aappling-usgs: xml2 is actually already a required package in vizlab (https://github.com/USGS-VIZLAB/vizlab/blob/master/DESCRIPTION#L30), so you can just call the `xml2::read_xml` with the double-colon syntax, as you're already doing below, and leave out the check for whether xml2 is available.

R/readData.R
```diff
@@ -207,6 +207,50 @@ readData.json <- function(viz) {
...
+  return(data.out)
+}
+
+#' \code{readData.svg_map} reads svg
+#' 
+#' @rdname readData
+#' @export
+readData.svg <- function(viz){
+  required <- c('location') 
+  checkRequired(viz, required)
```
aappling-usgs: :+1:

R/readData.R
```diff
@@ -207,6 +207,50 @@ readData.json <- function(viz) {
   return(data)
 }
 
+#' \code{readData.shp} reads shapefiles 
+#' 
+#' assumes only one .shp file in zip folder
+#' 
+#' @rdname readData
+#' @export
+readData.shp <- function(viz){
```
aappling-usgs: agreed that `readData.svg` is a better name for this than `readData.svg_map`

R/readData.R
```diff
@@ -207,6 +207,50 @@ readData.json <- function(viz) {
...
+#' 
+#' @rdname readData
+#' @export
+readData.svg <- function(viz){
+  required <- c('location') 
+  checkRequired(viz, required)
+  
+  has_xml2 <- suppressMessages(suppressWarnings(require(xml2)))
+  if(!has_xml2) { 
+    stop("xml2 package is required to read svg files (mimetype='image/svg+xml', reader='svg')")
```
aappling-usgs: the default mimetypes file routes two different mimetypes, 'image/svg' and 'image/svg+xml', to this reader (https://github.com/USGS-VIZLAB/vizlab/blob/67e4a89cb839c9663d8533de996de4e757c5eb76/inst/mimetypes.default.yaml#L14-L16). so maybe change this error message to acknowledge both?
```
stop("xml2 package is required to read svg files (mimetype='image/svg' or 'image/svg+xml', reader='svg')")
```

R/readData.R
aappling-usgs: i believe i fixed the jsonlite issue, which was in readData.json, in my currently open PR (https://github.com/USGS-VIZLAB/vizlab/pull/276/commits/82ae5df7e32925a77fbc4e7f5c331b87f75318c2#diff-1f1c30f18d0262fd78b81dec3207c046R198). so you can skip that one, maybe do another upstream pull & push to this branch after my PR is merged

R/readData.R
jzwart: I can do that although this error was for checking if xml2 package was loaded and it is already a required package in vizlab as you point out above so I was just going to get rid of error message.

aappling-usgs: oh, yep, good point

aappling-usgs:
Hey @jzwart , what do you think about also creating a test for each of these readers? Up to you whether you want to do that in this PR or in a subsequent one. 

You could add an example dataset of each format in https://github.com/USGS-VIZLAB/vizlab/tree/master/inst/testviz/data and the fetch block of https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/testviz/viz.yaml. The [default fetcher](https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/viz.defaults.yaml) is `file`, which should allow you not to write a fetcher for those examples. Then in https://github.com/USGS-VIZLAB/vizlab/blob/master/tests/testthat/tests-readData.R you could add a call to readData() for each item and confirm that it yields the expected R object.

jzwart:
Hey @aappling-usgs, yeah I can add checks on this PR.

R/readData.R
```diff
@@ -207,6 +207,45 @@ readData.json <- function(viz) {
...
+  }
+  
+  shp.path <- file.path(tempdir(), 'tmp')
+  if (!dir.exists(shp.path)){
+    dir.create(shp.path)
+  }
+  
+  unzip(viz[['location']], exdir = shp.path)
+  layer <- tools::file_path_sans_ext(list.files(shp.path, pattern='*.shp'))[1]
+  data.out = readOGR(shp.path, layer=layer)
```
jzwart: I'm testing readers with example files and I cannot open shapefile if `readOGR` is written as
`data.out = readOGR(shp.path, layer=layer)` 
but I can read in shapefile if `readOGR` is written as
`data.out = readOGR(file.path(shp.path, paste(layer, '.shp', sep='')))` 
where the dsn is pointing to the shapefile rather than a folder with associated layer. Any thoughts on what I'm doing wrong?

tests/testthat/tests-readData.R
```diff
@@ -14,11 +14,13 @@ test_that("readData works as expected", {
   mayfly <<- readData(viz[['depends']][1])
   carData <<- readData(viz[['depends']][2])
   plotData <<- readData(viz[['depends']][3])
+  svg <<- readData(viz[['depends']][4])
   expect_is(mayfly, 'data.frame')
   expect_is(carData, 'list')
   expect_is(carData[[1]], 'data.frame')
   expect_is(carData[[2]], 'data.frame')
   expect_is(plotData, 'list')
+  expect_is(svg, c('xml_document','xml_node'))
```
aappling-usgs: great!

tests/testthat/tests-readData.R
```diff
@@ -4,7 +4,7 @@ oldwd <- getwd()
 #these tests need to use the test viz
 testtmp <- setup(copyTestViz=TRUE)
 viz <- as.viz(list(
-  depends = c("MayflyNymph", "carData","plot-info")
+  depends = c("MayflyNymph", "carData", "plot-info", "exampleSVG")
```
aappling-usgs: i know i shouldn't care so much, but your attention to consistent spacing warms my heart =)

.travis.yml
```diff
@@ -16,6 +16,12 @@ r_github_packages:
  - jimhester/covr
  - USGS-R/grithub
 
+before_script:
+  - sudo apt-get update
+  - sudo apt-get install -y gdal-bin libgdal-dev libgdal1-dev netcdf-bin libproj-dev
+  - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
+  - chmod 755 ./travis-tool.sh
```
aappling-usgs: lines 22 and 23 look to me like old code, from the days before travis was good at R. Maybe we could get away without them?

.travis.yml
```diff
@@ -16,6 +16,12 @@ r_github_packages:
  - jimhester/covr
  - USGS-R/grithub
 
+before_script:
+  - sudo apt-get update
+  - sudo apt-get install -y gdal-bin libgdal-dev libgdal1-dev netcdf-bin libproj-dev
```
aappling-usgs: Is `netcdf-bin` strictly necessary for gdal? @jread-usgs , do you know?

aappling-usgs:
It looks like there may also be a newer and simpler way to install using `apt-get`: https://github.com/craigcitro/r-travis/issues/66#issuecomment-233969715  , e.g.:
```yml
addons:
  apt:
    packages:
      - libgdal-dev
      - libproj-dev
```

jzwart:
I'm a bit stuck. `The command "eval sudo apt-get update -qq " failed.` I tried adding `dist: trusty` to travis.yml but that didn't help.

aappling-usgs:
I'm not sure why the update wasn't working either, but try that `addons` syntax above :point_up: - maybe travis put some support behind that syntax that isn't available within `before_scripts`.

jzwart:
ok and get rid of 
```
before_script:
  - sudo apt-get update`
  - sudo apt-get install -y gdal-bin libgdal-dev libgdal1-dev netcdf-bin libproj-dev
  - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
  - chmod 755 ./travis-tool.sh
```

aappling-usgs:
yes, just make sure you get the right libs listed (the example i shared probably isn't complete)

jzwart:
still same error. I have to run but will be back on this later in afternoon

aappling-usgs:
looks like other people have been seeing the same error this week: https://github.com/travis-ci/travis-ci/issues/8581

aappling-usgs:
maybe try this? https://github.com/GeoscienceAustralia/mtpy2/commit/ed4292502e84113bb4ec5878e6377287ac7b605b

aappling-usgs:
https://www.traviscistatus.com/incidents/77q4ts757mzs

jzwart:
Yup!

PR #293, opened 2017-10-17
keeping up with master branch

aappling-usgs:


aappling-usgs:
merging myself only because this PR is entirely made of old commits, just merging them from the `master` branch into the `remake` branch

PR #295, opened 2017-10-17
Tooltip js function that doesn't get clipped top/right/left

jread-usgs:
ezgif com-crop

gif from how this works on the [`example` viz](https://github.com/USGS-VIZLAB/example/pull/22)


inst/js/tooltip.js
```diff
@@ -11,53 +11,86 @@ function hovertext(text, evt){
       clearTimeout(hoverTimer); //stop when off area
```
mbucknell: I'm a little leery of using global variables for hoverTimer and hoverDelay. I think you could potentially run into places where two hover events on different svgs are active and then there would be some unexpected behavior. It would take some reworking to use something like a module pattern to isolate hover events on different svgs.  Also is there a case where evt would be undefined? I'm assuming that this would be used as an event handler for an svg element.

After looking farther down, it looks like this is used primarily to have a google event registered. It might make more sense to have a google event at the start of a hover for an SVG and then one when the hover is dismissed.

inst/js/tooltip.js
```diff
@@ -11,53 +11,86 @@ function hovertext(text, evt){
...
-      " v"+(textHeight+textBuffer)+"Z");
+    
+    tooltipText.attr("x", tooltipX);
+    if (tipPointer.y < - (tipYoffset)){ 
+      // is a rectangle w/ no tip
+      
+      tooltipPath.attr("d", 
+        "M" + (tooltipX - halfLength - textBuffer) + "," + svgTopBound +
+        " H" + (tooltipX + halfLength + textBuffer) +
+        " v" + (textHeight+textBuffer) +
```
mbucknell: Nit. Better style to have leading/trailing space around operators like '+'

inst/js/tooltip.js
jread-usgs: Yeah, I think we could/should think about how these events should be handled and where they live. This is the part of the code I didn't touch for this work. 
I think the issue w/ sending an event when the hover starts is that it will exceed google's max number of events per second, because sliding the mouse across multiple elements would trigger events for all. The timer was implemented to only register a hover when someone hovered for a set amount of time. But again, I am not very familiar w/ this part of the code...

PR #299, opened 2017-10-17
Gtag

ldecicco-USGS:
@wdwatkins ....maybe we can look at this together?

### DIFF
--- a/inst/templates/footer.mustache
+++ b/inst/templates/footer.mustache
@@ -10,7 +10,7 @@
     </div>
 
     <div id="github-contact">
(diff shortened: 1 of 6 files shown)


inst/templates/footer.mustache
```diff
@@ -50,7 +50,7 @@
           <div class="relatedItemContainer">
           	{{#blogs}}
           		<div class="relatedItem">
-          		  <a href="{{url}}" target="_blank">
+          		  <a href="{{url}}" target="_blank" onclick="gtag('event','outbound', 'clicked on related blog')">
```
aappling-usgs: good idea to add this

PR #301, opened 2017-10-17
Embed

ldecicco-USGS:
I haven't been able to get this to work 100%


inst/templates/socialMedia.mustache
```diff
@@ -1,17 +1,24 @@
 <div id="socialHolder">
-  <a href="#" target="_blank">
+  <a href="{{githubLink}}" target="_blank">
```
ldecicco-USGS: @mwernimont any idea why this doesn't seem to do anything? When I build with this PR, I still get in source:
image

(the blank href being the annoying part)

mwernimont: Does it need 3 brackets? {{{}}} I'm just spit balling here, but will look at the rest of the code quick

mwernimont:
@ldecicco-USGS what does the yaml look like for {{githublink}} etc... We may just be trying attack this wrong.

ldecicco-USGS:
I think it's finally working!

image


Exception is some style, but I think now smarter design/css eyes can take it from here.

inst/resource.library.yaml
```diff
@@ -58,6 +58,7 @@
     facebook-logo: lib-facebook-logo
     embed-logo: lib-embed-logo
     css: lib-social-media-css
+    social-links:
```
aappling-usgs: does this need a resource after the colon?

ldecicco-USGS: in fact, leaving it blank I think is important

Issue #292, opened 2017-10-16
trouble with remake sources


aappling-usgs:
For issue #246 and PR #281 , the scripts I list as remake `sources` don't seem to be getting sourced. I'll keep working on it but would also welcome help - can y'all reproduce and/or diagnose this issue, @jread-usgs and others? 

Here's a simple remake.yml that I just created in my copy of USGS-VIZLAB/example:
```yml
packages:
  - vizlab
sources:
  - scripts/fetch/cars.R
targets:
  all:
    depends:
      - 'cache/cars.csv'
  'cache/cars.csv':
    command: fetch(I('cars-data'))
```
when I run `remake::make()` in a fresh R session (after deleting the .remake folder if it exists), I get this:
```r
> remake::make()
[  LOAD ] 
[  READ ]                |  # loading sources
<  MAKE > all
[ BUILD ] cache/cars.csv |  fetch("cars-data")
[  READ ]                |  # loading packages
Restoring previous version of cache/cars.csv
Error in UseMethod("fetch") : 
  no applicable method for 'fetch' applied to an object of class "c('cars', 'fetcher', 'fetch', 'viz')"
```
but if I then source the script manually and try again, everything works much better:
```r
> source('scripts/fetch/cars.R')
> remake::make()
<  MAKE > all
[ BUILD ] cache/cars.csv |  fetch("cars-data")
[ ----- ] all
```

This same pattern holds even if I copy scripts/fetch/cars.R into the top-level directory and change `sources` to simply `cars.R`, and even if I declare it in one line: `sources: cars.R`.

I'm up to date on `remake`:
```r
> packageVersion('remake')
[1] ‘0.3.0’
> devtools::install_github('richfitz/remake')
Using GitHub PAT from envvar GITHUB_PAT
Skipping install of 'remake' from a github remote, the SHA1 (e29028b5) has not changed since last install.
  Use `force = TRUE` to force installation
```

Any ideas why this is happening or how to make my `sources:` get `source()`d?

aappling-usgs:
Narrowed it down a bit - I think the problem is that `remake` is so selective about what it loads that it's not bothering to load `fetch.cars` when the command just says `fetch()`.

In a new, nearly empty directory, I have these three files:

### cars.R
```r
fetch.cars <- function(viz) {
  viz <- as.viz(viz)
  id <- viz[['id']]
  location <- viz[['location']]
  write.csv(x=cars, file=location, row.names=FALSE)
}

[...]

PR #305, opened 2017-10-18
Refine first-cut remake implementation

### LINKED ISSUE #246: Switch to remake?
This would be a fair amount of work, but would help with...
* dependency diagrams (https://github.com/USGS-VIZLAB/vizlab/issues/204)
* `make` configuration management (https://github.com/USGS-VIZLAB/vizlab/issues/209, https://github.com/USGS-VIZLAB/vizlab/issues/183, https://github.com/USGS-VIZLAB/vizlab/issues/117)
* debugging with breakpoints
* consistency with team's data pipeline activity
* speed (i think...) for many small R scripts, e.g., might enable a faster implementation of https://github.com/USGS-VIZLAB/vizlab/issues/193
* package dependencies (#91)

HOWEVER, fetchTimestamp might behave really differently under a remake solution. Figure out a plan for this before putting too much energy into a switch.

aappling-usgs:
At this point, using the `remake` branch of `vizlab` (including this PR) you should be able to make and more or less build the `example` repo using
```r
library(vizlab)
vizmake()
```

so this PR fixes #246 .

Things I did in this PR:
* add job blocks so you can `vizmake('Fetch')`, `vizmake('Visualize')`, etc. to build just one phase at a time
* add explicit dependencies of each item on its declared scripts (or functions within scripts). if there are scripts declared for an item, `prepSources` first combines and/or subsets those into a single item-specific script file, stored in vizlab/remake/scripts. then the viz item depends more specifically on that file. `prepSources` also removes comments and whitespace, so you can add those without forcing a rebuild of an item. all of this is necessary because of #292, and it's a follow-on to #294 where i took a good stab at fixing that problem but didn't succeed in re-coupling remake targets to their script dependencies.


jread-usgs:
very nice :+1: `vizmake('Fetch')`

PR #310, opened 2017-10-19
remove autoargs (fixes #303)

### LINKED ISSUE #303: delete getAutoargs
I'm pretty sure we don't use it any more. The only places where it gets called are within other Autoargs functions and in two commented-out lines (in fetch.R and visualize.R)

aappling-usgs:
autoargs are an obsolete approach to pulling in information from viz items; now we use readDepends and direct references to argument names. This PR just cleans out the old, unused code.


ldecicco-USGS:
Seems pretty straightforward to me. I'll wait for the checks to finish before I merge

PR #311, opened 2017-10-20
adding fetch.usgs_watermark 

### LINKED ISSUE #122: add USGS watermark SVG to the package
image

for matthew. 

Like the favico, it might be nice to have this in the package (`inst/extdata/` or `sysdata.rda`?) so it is easy to pull in

jzwart:
fixes #122. I wasn't sure where to put a tester, but took a first crack at it. Still need to add alignment in publish which may be opened as a different issue or could be added to this PR.


R/fetchTimestamp.R
```diff
@@ -103,6 +103,37 @@ fetchTimestamp.file <- function(viz) {
...
 }
 
+#' `fetchTimestamp.usgs_watermark` creates a timestamp file once, with file metadata and
+#' file contents that both match the timestamp of the viz data file. If the
+#' timestamp file already exists, it is only modified if the data file is
+#' updated.
+#' 
+#' @rdname fetchTimestamp
+#' @export
+fetchTimestamp.usgs_watermark <- function(viz) {
```
aappling-usgs: if this function is truly identical to `fetchTimestamp.file` (it is, right?) then you can actually just define it as
```r
fetchTimestamp.usgs_watermark <- fetchTiimestamp.file
```
and call it a day

PR #314, opened 2017-10-26
Fixes #308

### LINKED ISSUE #308: Get better thumbnail requirements
Right now we've got the `checkThumbCompliance` checking single exact dimensions, when in actuality there's a range. Presumably, that is true for twitter too.

"Use images that are at least 1200 x 630 pixels for the best display on high resolution devices. At the minimum, you should use images that are 600 x 315 pixels to display link page posts with larger images. Images can be up to 8MB in size."

https://developers.facebook.com/docs/sharing/best-practices
https://developers.facebook.com/docs/sharing/insights

Also...facebook analytics?
https://developers.facebook.com/products/analytics/quickstarts/web

ldecicco-USGS:


ldecicco-USGS:
So this is a small subset of what I'm working on with thumbnails, but it's it's a digestable chunk that should carry over no matter how we clean up the rest of the thumbnail business.

R/publish.R
```diff
@@ -405,43 +404,108 @@ publish.thumbnail <- function(viz){
...
+                               exactWidth = exactWidth)
   #send to other publishers if all ok
   viz <- NextMethod()
   viz[['url']] <- pastePaths(getVizURL(), viz[['relpath']])#need to add slash between?
   viz[['width']] <- dims[['width']]
   viz[['height']] <- dims[['height']]
 }
 
+
+convb <- function(x){
```
aappling-usgs: looks fancy. could you add a short comment explaining what this function does?

R/publish.R
```diff
@@ -405,43 +404,108 @@ publish.thumbnail <- function(viz){
...
+    maxHeight <- 4096
+    maxWidth <- 4096
+    minHeight <- 157
+    minWidth <- 300
   } else { #landing
     maxSize <- 1048576
-    checkHeight <- 400
-    checkWidth <- 400
+    exactHeight <- 400
+    exactWidth <- 400
```
aappling-usgs: these are better names

R/publish.R
```diff
@@ -405,43 +404,108 @@ publish.thumbnail <- function(viz){
...
+  if(width > maxWidth | height > maxHeight) {
+    stop(paste("Thumbnail", file, "is too big"))
   }
-
+  
+  if(width < minWidth | height < minHeight) {
+    stop(paste("Thumbnail", file, "is too small"))
+  }
+  
+  if(width == exactWidth & height == exactHeight) {
```
aappling-usgs: i like these more specific error messages. for the `if()` tests, i think you could still use `||` and `&&` instead of the single-character versions, couldn't you?

ldecicco-USGS: I think yes for `&&`....not sure on the "or's". Probably? Yes...I'm definitely leaning, most probably yes

aappling-usgs: me too. absomostly.

Issue #250, opened 2017-10-03
require thumbnails in `publish` section only


aappling-usgs:
move width and height and URL information to vizlab internals so that we only need to have thumbnail items in the `publish` section, not also in `info`. make sure this works with the existing functionality for creating thumbnail metadata (`<meta property="og:` etc. in viz page) and with `checkThumbCompliance()`

aappling-usgs:
closed by #319...will give github some time to consider closing this automatically, then will close manually if needed

PR #319, opened 2017-10-31
Thumbnail

ldecicco-USGS:


ldecicco-USGS:
Fixes #250 

Bonus: dev version of `publishLandingPage`

image


Not sure if there's a better way to get the thumbnails on the unpublished viz's, but it's a step in the right direction

R/publish.R
```diff
@@ -392,18 +446,71 @@ publish.landing <- function(viz){
...
+    file.info(viz[['location']])
+  },
+  error=function(cond){
+    return(file.info(system.file("testviz/images/landing-thumb.png", package = "vizlab")))
+  })
+  for(thumbType in unique(viz[['thumbType']])){
+    dims <- checkThumbCompliance(width, height, file_size$size, thumbType)
+  }
+
+  viz[['url']] <- pastePaths(getVizURL(), viz[['relpath']])#need to add slash between?
```
aappling-usgs: the answer to the comment is no.

R/publish.R
```diff
@@ -392,18 +446,71 @@ publish.landing <- function(viz){
...
+}
+
+#' helper to check thumbnail compliance
+#' @importFrom imager load.image width height
+#' @param width numeric pixal width of image
+#' @param height numeric pixal height of image
+#' @param size numeric file size in bytes
+#' @param thumbType char Type of thumbnail, could be "facebook", "twitter", "landing", "main"
+checkThumbCompliance <- function(width, height, size, thumbType){
+  match.arg(thumbType, c("facebook","twitter","main","landing"))
```
aappling-usgs: :+1: but maybe 
```
thumbType <- match.arg(thumbType, c("facebook","twitter","main","landing"))
```
because match.arg won't complain if `thumbType` is `'face'` (and that's intended to be a feature, not a bug), but `thumbType` won't be translated to `facebook` without the assignment.

Alternatively, if you want to take a harder line:
```
stopifnot(all(thumbType %in% c("facebook","twitter","main","landing"))
```

R/publish.R
```diff
@@ -51,6 +52,60 @@ publish.page <- function(viz) {
   return(relativePath(file))
 }
 
+#' Get thumbnail info for sematics
+#' @param context list, should be just info section
+#' @param thumbnails list, taken from full context
+update_thumbnails <- function(context, thumbnails){
```
aappling-usgs: seems kinda silly to worry about it b/c we're doing snake_case for actual vizzies, but vizlab follows camelCase...

R/publish.R
```diff
@@ -353,19 +406,20 @@ publish.social <- function(viz) {
...
 #'
 #' @rdname publish
+#' @param dev logical whether or not to publish development version of landing page
 #' @export
-publish.landing <- function(viz){
+publish.landing <- function(viz, dev, ...){
 
   repos <- getRepoNames(viz[['org']])
-  viz_info <- lapply(repos, getVizInfo, org=viz[['org']])
+  viz_info <- lapply(repos, getVizInfo, org=viz[['org']], dev)
```
aappling-usgs: last arg could be `viz[['dev']]` instead of `dev` if following suggestion below at https://github.com/USGS-VIZLAB/vizlab/pull/319/files#diff-58996d4f2f3624e968608c612deab004R15

R/publishLandingPage.R
```diff
@@ -2,16 +2,17 @@
...
+publishLandingPage <- function(dev = FALSE){
 
   file.copy(from=system.file('landing', package="vizlab"), to=getwd(),
             recursive = TRUE, overwrite = TRUE)
   oldwd <- getwd()
   setwd('landing')
   on.exit(setwd(oldwd))
   
-  publish('landing')
+  publish('landing', dev=dev)
```
aappling-usgs: Having a dev version of publish is an awesome idea. I'd prefer a slightly different implementation (though am open to discussing) - specifically, Jordan Walker worked so hard to make fetch/process/visualize/publish be single-argument functions that I'd like to us to try to preserve that. 

One way to do this would be to replace this line with:
```r
    viz <- as.viz('landing')
    viz <- as.publisher(viz)
    viz$dev <- dev
    publish(viz)
```
and then to have 
```r
viz_info <- lapply(repos, getVizInfo, org=viz[['org']], viz[['dev']])
```
here: https://github.com/USGS-VIZLAB/vizlab/pull/319/files#diff-b2d70d609ac0e9c4f742e4a599d45f19R422

R/publishLandingPage.R
```diff
@@ -40,26 +41,35 @@ getVizYamlUrl <- function(org, repo){
...
+        return()
+      } 
+    } else {
       return()
-    } 
+    }
   } else {
-    return()
+    publish_date <- as.Date(viz.yaml$info$`publish-date`)
+    is_published <- TRUE
```
aappling-usgs: cool!

R/publish.R
```diff
@@ -353,19 +406,20 @@ publish.social <- function(viz) {
...
   names(viz_info) <- repos
   # rm null
   viz_info <- viz_info[!sapply(viz_info, is.null)]
```
aappling-usgs: I don't think this is the PR or sprint where we should do this, but I'm wondering: Is it possible that `publishLandingPage` could eventually call `vizmake()` after creating and changing into the landing directory? That would allow us to use remake to publish the thumbnails and other resources rather than having `publish` calls sprinkled throughout the `publish` functions...what are your thoughts while this is all fresh in your mind, @ldecicco-USGS ?

inst/testviz/viz.yaml
```diff
@@ -171,16 +173,26 @@ publish:
...
     mimetype: image/png
-    export: TRUE
   -
     id: landing-thumb
     location: images/landing-thumb.png
     publisher: thumbnail
-    for: landing
+    thumbType: landing
+    title: "Landing thumbnail"
+    alttext: "Alt text thumbnail"
```
aappling-usgs: if/when we write some formal tests for this functionality, what are the important features to capture? thumbnail pngs should get copied into target? thumbnails should appear in target header?

R/publishLandingPage.R
ldecicco-USGS: Oh! I didn't realize leaving out the `...`'s was by design. I actually assumed it was an oversight to not have them there. OK, I'll make the change

inst/testviz/viz.yaml
ldecicco-USGS: There are some tests in the publish section:
```
test_that("Thumb publisher works", {
  expect_error(publish("facebook-thumb")) #incorrect dimensions
  publish('landing-thumb')
  expect_true(file.exists('target/images/landing-thumb.png'))
})
```

aappling-usgs: cool, and oops, i forgot that tests might already exist and was only looking at files changed in this PR. silly me - you'd think i'd remember given that my checks of some branches have been breaking on this test for a day or two =).

R/publish.R
```diff
@@ -401,9 +453,63 @@ publish.template <- function(viz) {
...
+
+#' helper to check thumbnail compliance
+#' @importFrom imager load.image width height
+#' @param width numeric pixal width of image
+#' @param height numeric pixal height of image
+#' @param size numeric file size in bytes
+#' @param thumbType char Type of thumbnail, could be "facebook", "twitter", "landing", "main"
+checkThumbCompliance <- function(width, height, size, thumbType){
+  
+  stopifnot(all(thumbType %in% c("facebook","twitter","main","landing")))
```
ldecicco-USGS: you've got me questioning if I should replace all my `match.arg` calls I have spewed about my package coding

aappling-usgs: Hmm. I do think partial matching (e.g., 'face' to 'facebook') is sort of nice, and is consistent with behavior in other base R functions. And you do get an error on an ambiguous match with `match.arg`, so it's not super dangerous that way. But it's true that someone could have passed in a thumbType of 'face' and then seen the surprising behavior that the `if('facebook')` code block never got executed. In general I quite like the `x <- match.arg(x, ...)` pattern - that one might be appropriate in other packages where you've used `match.arg`, while in vizlab our reliance on syntax makes this a good place for a hard `stopifnot`.

Issue #287, opened 2017-10-16
port the dependencies on non-file viz.yaml options to remake.yaml


aappling-usgs:
PR #281 introduces `createRemakefile()`, which i hope will let us soon replace `make` with `remake`. but #281 doesn't yet include targets for the viz.yaml non-file options, which were introduced for `make` in #193. We need to pull these over into our `remake` system.

In this process, let's also look for opportunities to be more efficient with `remake` than we were with `make`, if possible, because our system from #193 ended up being sorta slow in practice. This probably means using R objects rather than files to store the viz.yaml option chunk for each viz item. It might also involve pre-processing the viz.yaml into an R list of all the chunks before separating into one chunk per object?

aappling-usgs:
closed by #321

Issue #245, opened 2017-10-03
Make publish step take advantage of make


aappling-usgs:
It seems not to do this. Surely it could.

PR #321, opened 2017-10-31
remake items now depend on viz.yaml arguments

### LINKED ISSUE #215: tell `make` the provenance of resources from vizlab
This wasn't a problem until vizlab version 0.2.x, which introduces dependencies on the viz.yaml arguments for each item. But now there are errors when you try to `make publish` the example repo. The reason is that one of the dependencies of a publish item (`footerCSS: "lib-footer-css"` under `index`) is a file available in the vizlab package (vizlab/inst/resource.library.yaml, https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/resource.library.yaml#L213-L217). It gets copied over somewhere, but not by `make`, so `make` has no guarantee that it will show up.

Possible solution: after #88 has been finished, we could make a new makefile `resources.make` that is responsible for pulling in files on which publish steps depend (pull from vizlab to the current vizzie). Given the expected approach to #88, we expect that then `resources` and `publish` would be able to see one another's targets, so `publish` could proceed comfortably.

aappling-usgs:
main goal was #287 (mainly in ed9c7cb), but this PR also
* confirms the resolution of #207 (because it was already done, just needed checking with #287 resolved) and
* resolves #215 (because i couldn't get tests to run without finishing this). leaves open #245, which i thought i was on the way to fixing in this sprint but which turns out to be a different beast from what i first thought.


R/remake.R
```diff
@@ -0,0 +1,17 @@
...
+#' that those items are rebuilt when their viz arguments change. The returned
+#' list excludes resources, which aren't listed explicitly in the viz.yaml and
+#' therefore can't have arguments.
+#' @param viz.yaml the name of the viz.yaml. should always be 'viz.yaml'.
+#'   Included as an argument so remake can make the dependency on the viz.yaml
+#'   explicit.
+#' @export
+collectItemArguments <- function(viz.yaml) {
+  stopifnot(viz.yaml == 'viz.yaml')
+  cis <- getContentInfos(block=c("parameter", "fetch", "process", "visualize", "publish"))
```
ldecicco-USGS: maybe a new variable name here? I was initially very confussed

R/vizmake.R
```diff
@@ -16,9 +16,15 @@ vizmake <- function(target_names=NULL, ...) {
   # check whether the right packages are installed
   checkVizPackages()
   
-  # decide whether createRemakefile needs to be rerun
-  if(!file.exists('remake.yaml') || file.mtime('remake.yaml') < file.mtime('viz.yaml')) {
-    createRemakefile()
+  # run createRemakefile and createDirectories if needed
+  if(!file.exists('remake.yaml') ||
+     file.mtime('remake.yaml') < file.mtime('viz.yaml') ||
+     file.mtime('remake.yaml') < file.mtime(system.file('DESCRIPTION', package='vizlab'))) {
```
ldecicco-USGS: :1st_place_medal:

inst/resource.library.yaml
```diff
@@ -106,6 +106,10 @@
   publisher: template
   mimetype: text/mustache
   depends:
+    brandcomment: brandcomment
```
ldecicco-USGS: Yeah, I went back and forth on if we should include this stuff in embed. I'm fine either way.

aappling-usgs: oh, i see. i just added them to `depends` because they were present in `context` and `partials`...did i misunderstand how that works?

ldecicco-USGS: you didn't mis-understand...I just forgot to EITHER delete them from the context (partials) OR put them in the depends. So...I'm happy you put it in there. If we decide we're bloating the embed, we can take it out.

tests/testthat/tests-timestamps.R
```diff
@@ -15,18 +15,15 @@ test_that("alwaysCurrent doesn't get rebuilt unless missing", {
   suppressWarnings(expect_message(vizmake('mayfly_nymph'), '[ BUILD ] data/mayfly_nymph.csv', fixed=TRUE))
   
   # after that: doesn't fetch
-  suppressWarnings(expect_message(vizmake('mayfly_nymph'), '[ ----- ] mayfly_nymph', fixed=TRUE))
+  suppressWarnings(expect_message(vizmake('mayfly_nymph'), '[    OK ] mayfly_nymph', fixed=TRUE))
```
ldecicco-USGS: total side note.....I'd often wanted a function that's similar to this for general use (suppress this one particular message, show the rest)

aappling-usgs: me too! gets really messy really quickly with calling handlers and all that

PR #322, opened 2017-11-02
Merge remake branch into master

aappling-usgs:
Tests pass, with some warnings I think are best fixed in a different PR.


aappling-usgs:
We keep seeing this decrease in coverage as measured by coveralls. It's disappointing, but I think we can accept the change because we were adding tests left and right all sprint - I blame the metric rather than our actions, at least for now.

aappling-usgs:
@ldecicco-USGS - huge number of commits in this PR, but only one that's new: https://github.com/USGS-VIZLAB/vizlab/pull/322/commits/8dd3f5ed0e4853d7a88b57487ecb976592be6e48. Could you just review that one?

Issue #327, opened 2018-01-10
devtools::test warnings and errors


aappling-usgs:
see travis build at https://github.com/USGS-VIZLAB/vizlab/pull/326, but the same issues are currently coming up for the master branch. i'm getting 9 warnings and 13 errors locally on both master and this PR. i'm guessing the issues are a combination of (1) sciencebase's new prohibition of public items from personal accounts, (2) formatting changes in remake's output, and maybe (3) something related to #325.

```r
==> devtools::test()

Loading vizlab
Loading required package: testthat
Testing vizlab
analytics: ..
checkVizPackages: .....
embed: .
markdown: .........
publish: ..............
publish viz landing page: WW....
readData: ........
readDepends: .....
remake: W.................
fetchTimestamps working: .12.34567.89.ab.....WWcdWW..WW...
util: ....
viz-setup: Creating directories ...
Creating viz.yaml ...
..........
testviz: Creating .gitignore
Copying LICENSE
........
sciencebase: .
templates: ...

Warnings -----------------------------------------------------------------------
1. publishLandingPage works (@tests-publishLandingPage.R#8) - NAs introduced by coercion: . is not a real

2. publishLandingPage works (@tests-publishLandingPage.R#8) - NAs introduced by coercion: . is not a real

3. createRemakefile makes remake.yaml (@tests-remake.R#8) - cannot find dependency '' of viz item 'social-media-template' so cannot include the dependency in remake.yaml

4. sciencebase works (@tests-timestamps.R#114) - Item not found for ID=575d839ee4b04f417c2a03fe. Either the item does not exist or the item is secured and requires authentication to access.

5. sciencebase works (@tests-timestamps.R#114) - could not retrieve timestamp from ScienceBase; local timestamp information will be left as-is

6. sciencebase works (@tests-timestamps.R#122) - Item not found for ID=575d839ee4b04f417c2a03fe. Either the item does not exist or the item is secured and requires authentication to access.

7. sciencebase works (@tests-timestamps.R#122) - could not retrieve timestamp from ScienceBase; local timestamp information will be left as-is

8. sciencebase works (@tests-timestamps.R#133) - Item not found for ID=575d839ee4b04f417c2a03fe. Either the item does not exist or the item is secured and requires authentication to access.

9. sciencebase works (@tests-timestamps.R#133) - could not retrieve timestamp from ScienceBase; local timestamp information will be left as-is

Failed -------------------------------------------------------------------------
[...]

PR #328, opened 2018-01-14
publish.json, publish.tabular, and simpler export, buildContext, and expandDependencies

### LINKED ISSUE #263: publish.json method
Inspired by #41 and the need for a publish.json method for Hurricane Harvey (https://github.com/USGS-VIZLAB/hurricane-harvey/blob/master/scripts/publish/json.R) and Matthew before that.

Comments from that viz on publishing a json data file: https://github.com/USGS-VIZLAB/hurricane-harvey/blob/acad75321841d9bcdde8f10ef43cb6fc2f4f87df/viz.yaml#L279-L283
```
    comment: "TOTAL HACK, writing into TARGET. How the heck do you get the other json in there?"
    comment-reply: >-
      export true allows it to be exported to target, 
      mimetype javascript is a hack to use a publisher that works, we should be able
      to either add a json publisher or, general data publisher so this ends up in data/times.json.
```

The publisher at https://github.com/USGS-VIZLAB/hurricane-harvey/blob/master/scripts/publish/json.R is a good start, but should probably include a call to `export` which I hope will resolve the concerns raised in the viz.yamls of Matthew and Harvey (see notes in #41). I think - but am not sure! - that once we get this right, the fetch step for a json file won't need to include an `export: true`, because the exporting can happen during `publish` instead. Whoever takes on this issue should look into that.

aappling-usgs:
This PR primarily resolves #263, which requests a publish.json method. I also added a publish.tabular method because both are needs for using d3 (or jquery) for more complex javascript-side data viz, and because the two methods are identical. 

The final paragraph in the description of issue #263 suggests that publish.json should include a call to `export`; this is accomplished by the call to `NextMethod`, which will resolve to `publish.resource`, which calls `export` and also (importantly!) copies the file into `target/data`.

Partly to make the file copying happen, and partly because I can't think of cases where file copying *shouldn't* happen, I've made the default in `export.resource` be for doExport to be TRUE. This allowed me to simplify the use and implementation of `doExport` throughout `export.R`. (I think there are further simplifications we could make to the `export` function family in the future, but this change is enough for one day.)

There were two places where defaults were getting set for the `export` parameter: in `doExport` and also in `inst/viz.defaults.yaml`, which affects how the viz.yaml is read in. I took out the `export` values in the latter file so that we could set values in just one place. I'm pretty sure the values set in viz.defaults.yaml were doing very little, or possibly even getting in the way.

While I was learning this area of the vizlab code, I had to look at `buildContext` and `expandDependencies` a bunch. I think they're more complex than necessary and am taking that gamble here by simplifying them. If I'm wrong and the reviewer of this PR also doesn't catch it, I expect we'll catch it soon enough when a bug comes up. You can blame me then (I even put my name in the code where I took out the functionality).

### DIFF
--- a/inst/mimetypes.default.yaml
+++ b/inst/mimetypes.default.yaml
@@ -25,5 +25,6 @@ js:
   - "text/javascript"
 json:
   - "application/json"
+  - "application/geo+json"
 css:
   - "text/css"
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: vizlab
 Type: Package
 Title: Utilities for building online data visualizations
-Version: 0.2.2.9008
-Date: 2018-01-09
+Version: 0.2.2.9009
+Date: 2018-01-13
 Author: Jordan Walker, Alison Appling, Lindsay Carr, Luke Winslow, Jordan Read, Laura DeCicco, Marty Wernimont
 Maintainer: Alison Appling <aappling@usgs.gov>
 Description: Provides utility functions to organize, develop, and publish
--- a/R/export.R
+++ b/R/export.R
@@ -103,14 +103,15 @@ export.css <- function(viz) {
   return(file)
 }
 
-#' Default export, defaults to not export
+#' Default export. Used to default to not export, but Alison couldn't think of
+#' when that would be useful so changed the default to TRUE.
 #'
 #' @rdname export
 #' @export
 export.resource <- function(viz) {
   location <- viz[['location']]
   file <- NULL
-  if (doExport(viz, FALSE)) {
+  if (doExport(viz, TRUE)) {
     file <- paste0(exportLocation(), "data/", basename(location))
   }
   return(file)
--- a/R/export.R
+++ b/R/export.R
@@ -28,7 +28,7 @@ export.page <- function(viz) {
   }
 
   file <- NULL
-  if (doExport(viz, default = TRUE)) {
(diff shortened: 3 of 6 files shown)


aappling-usgs:
@lindsaycarr, because you're working on vizstorm Monday, I'm requesting your review and merge on this PR.

@ldecicco-USGS, I'm also hoping that you will review the code post-merge sometime during the workweek because you wrote a harvey or matthew version of publish.json and may have some relevant experience with this set of functions. Maybe you can catch a mistake or two - my understanding of the publish phase is still a bit shaky.

aappling-usgs:
I believe the Travis failures are the same ones noted in the last PR, having to do with changes in ScienceBase and remake. Issue #327 documents these issues. I think we can take them on another day, in favor of getting vizstorm schnazzy for tech tues.

inst/mimetypes.default.yaml
```diff
@@ -25,5 +25,6 @@ js:
   - "text/javascript"
 json:
   - "application/json"
+  - "application/geo+json"
```
lindsaycarr: curious, why is this `geo+json` instead of just `geojson`?

aappling-usgs: not sure. all i know how to do with mimetypes is google them - they remain a mystery to me =). an earlier mimetype for geojson was `application/vnd.geo+json`, which only adds to the mystery IMO

Issue #334, opened 2018-03-01
testviz should build cleanly


aappling-usgs:
This would really help with testing, especially adding new tests. Modifications could be made to introduce specific errors in specific tests, which would allow those tests to check for expected responses.

PR #335, opened 2018-03-01
Unpublish

aappling-usgs:
New function = unpublish(). Use case: you deleted your target folder. Simply running vizmake() again won't recreate the folder. So instead, run
```r
unpublish()
vizmake()
```
and the folder should get rebuilt.

The vizlabPublish - remake collision is a problem if you delete your target folder and then try to remake, because things just won't republish. This is my solution: delete the remake database info about those publish items, so that they get rebuilt on the next call to vizmake, but don't delete the files because usually the nominal targets of the remake recipes that call publish() are actually files that exist elsewhere, so we really don't want to delete those files.

I tried to add tests but was partly stymied by the fact that the testviz doesn't build without errors. So I added those tests I could manage, added https://github.com/USGS-VIZLAB/vizlab/issues/334, and ran the commented-out tests below in water-use-15, where they do behave as I'd expect.


aappling-usgs:
let's see if https://github.com/USGS-VIZLAB/vizlab/pull/336 can fix the sciencebase errors on travis.

Issue #339, opened 2018-03-02
`get the code` link is broken


lindsaycarr:
Noted first in `how-to-gdp` and then again in `water-use-15`, so something must be wrong in the templating for this. See https://github.com/USGS-VIZLAB/how-to-gdp/issues/52 and https://github.com/USGS-VIZLAB/water-use-15/issues/60.

PR #341, opened 2018-03-05
Correctly getting the github url in there

ldecicco-USGS:


ldecicco-USGS:
Fixes #339

ldecicco-USGS:
FYI, the testthat failures I get are still the timestamp issues we've been seeing.

aappling-usgs:
Are you getting those testthat failures locally, Laura? Looks like we're now getting mixed messages from Travis - the [PRs have been building successfully](https://travis-ci.org/USGS-VIZLAB/vizlab/pull_requests) while the [merges have been timing out](https://travis-ci.org/USGS-VIZLAB/vizlab/builds). I don't understand how those classes of Travis tests differ, but this new pattern seems slightly better than everything failing...and the failures are no longer pointing directly to sciencebase timestamps, at least.... #327

ldecicco-USGS:
Yes, I was getting those locally

aappling-usgs:
Any chance you're now just getting `crayon` errors, not sciencebase timestamp ones? see https://github.com/USGS-VIZLAB/vizlab/issues/327#issuecomment-369628206 for what i think our current status is. and if you're getting something different from me, could you paste your current messages into that issue?

ldecicco-USGS:
Yup, that's it!

PR #343, opened 2018-03-07
css suggestions for vizlab that came from how-to-gdp reviewer

lindsaycarr:
Ramona from TX WSC viz group had good tips about some the html tags we were using, including how using `<h3>` in the main header "will seriously discount that for automated crawlers in favor of the `<h1>`". That has now been adjusted :)

PR #344, opened 2018-03-08
change the date only

aappling-usgs:
Testing whether it's just Travis today that's breaking or whether it's changes in #343 that are breaking


aappling-usgs:
No need to merge; I was just testing for contrast with #343.

PR #353, opened 2018-03-14
footer changes

mwernimont:
So making these changes for the gdp-how viz, looks to be fixing the water-use footer issues as well.

### DIFF
--- a/inst/css/footer.css
+++ b/inst/css/footer.css
@@ -33,35 +33,61 @@ footer a:active{
   padding:0 10px;
 }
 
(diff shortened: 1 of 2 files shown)


inst/css/footer.css
```diff
@@ -33,35 +33,61 @@ footer a:active{
   padding:0 10px;
 }
 
+#vizlabPromo{
```
aappling-usgs: could we do hyphen separation for `vizlab-promo`?

PR #355, opened 2018-03-15
work on crayon & grithub errors

aappling-usgs:
trying to get tests to pass in more places and more often.


aappling-usgs:
Shoot, this includes lindsay's code that i was trying to complement rather than copy. no good.

PR #359, opened 2018-03-16
css update title wrap + links in footer

lindsaycarr:
Still noticed some tiny details in how-to-gdp that would be useful for future vizzies. 

On mobile, the header words wrapped and disappeared:
image

This fixes that:

image

Desktop remains the same:
image

Also in the footer, the title "Related Visualizations" looked and acted like a link but didn't link to anything, so changed the template. Same for the title of the related vizzie, but I added the URL to that since it should link.

No noticeable appearance, but underlines on hover only work for the ones with links now.
image

### DIFF
--- a/inst/css/header.css
+++ b/inst/css/header.css
@@ -1,7 +1,7 @@
 #header{
 	background:rgb(0,51,102);
 	color:#fff;
-	height:32px;
+	height:auto;
 	padding:8px  0 0 10px;
 	border-bottom:5px solid rgb(210,210,210);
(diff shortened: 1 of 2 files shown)


aappling-usgs:
@mwernimont could you take a look and merge when ready? it's short.

mwernimont:
I would make the title smaller on mobile over making room for it to map.

lindsaycarr:
Resized font based on screen size. 

Desktop:
image

Mobile:
image

@mwernimont did I do that right?

mwernimont:
I like that more than wrapping!

PR #361, opened 2018-03-22
Fixes #360

### LINKED ISSUE #360: analytics block of fullPage template not rendering
This section: https://github.com/USGS-VIZLAB/vizlab/blob/master/inst/templates/fullPage.mustache#L21 of the `fullPage` template doesn't seem to be rendering, even with the `analytics-id` viz.yaml field filled in.  This is in `water-use-15`.

ldecicco-USGS:


ldecicco-USGS:
So, if a analytic id is used, the source will have:
```

<script type="application/javascript" src="https://www2.usgs.gov/scripts/analytics/usgs-analytics.js"></script>
 
  <script async src="https://www.googletagmanager.com/gtag/js?id=XXX2"></script>
  <script>
  window.dataLayer = window.dataLayer \|\| [];
  function gtag(){dataLayer.push(arguments)};
  gtag('js', new Date());
  
   gtag('config', 'XXX');
   </script>
```

And if not, none of that is there (tested with water use viz)

aappling-usgs:
What was the secret? Is it just that you can't open and close `{{#info}}` twice?

ldecicco-USGS:
That's my observation. I really need to learn the finer details of mustache templating stuff

PR #367, opened 2018-04-24
New secondary and main svg

mwernimont:
So changed svg-container to main svg as it has the code for the big fancy interactive SVG.  The secondary was needed to avoid them inheriting the main-svg width CSS on Desktop. 

I think this works for now, but maybe some more thought needs to go into this for fixing IE and having the ability to do as many fancy SVG's as we want as easily as we can.

### DIFF
--- a/inst/css/content.css
+++ b/inst/css/content.css
@@ -30,7 +30,7 @@ body {
 /*IE SVG Width Fix*/
 @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none){
     .main-svg svg{
-      width: 100%;
+      width:100%;
       height:600px;
     }
     
--- a/inst/css/content.css
+++ b/inst/css/content.css
@@ -29,19 +29,24 @@ body {
 
 /*IE SVG Width Fix*/
 @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none){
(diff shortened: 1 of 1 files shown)


mwernimont:
On a CSS change?  What did I do!?

mwernimont:
Error: package ‘devtools’ was installed by an R version with different internals; it needs to be reinstalled for use with this R version
Execution halted
The command "Rscript -e 'devtools::install_github(c("jimhester/covr", "richfitz/remake", "USGS-R/grithub"), build_vignettes = FALSE)'" failed and exited with 1 during .

mwernimont:
Got ya.  I couldn't imagine it had a problem with the changes. lol

aappling-usgs:
Yeah, same error twice =(. It's clearly not getting to the point of testing any of your changes, and I've been bitten by overconfidence that changes won't affect the build before...but in the interest of keeping on keeping on with water use, I'm going to merge. :crossed_fingers:

aappling-usgs:
Oh, crud, I forgot to ask you to increment the package version and require the new version in water-use-15 within these PRs i just closed =(.

mwernimont:
Ugh, that gonna be a big issue?  I'm trying to help not inhibit. :(

PR #369, opened 2018-04-25
custom GA dimensions in gtag config

wdwatkins:


wdwatkins:
This sets up custom dimensions dimensions that we will send with the GA events (I should have a PR soonish in the water-use viz that utilizes this).  Mostly I think this is mapping the names "timestamp" to the dimension indices, so we can reference the descriptive name instead of the index.

wdwatkins:
```
Error: package ‘devtools’ was installed by an R version with different internals; it needs to be reinstalled for use with this R version
```

Any idea what this is about?  Does travis have R 3.5 installed now or something?

aappling-usgs:
Travis: sigh. I think it might be this: https://github.com/travis-ci/travis-ci/issues/9540. Same thing was happening with Marty's also-very-small PR. Could you just confirm that `devtools::check()` works locally for you? Expecting 0 warnings, 0 errors, 1 note about use of `:::`.

PR #372, opened 2018-04-26
testing travis

ldecicco-USGS:


ldecicco-USGS:
OK, so I think I've fixed the devtools thing. I was reading "It is because you are using `sudo: required` and when that happens the devtools binary is installed, but the binaries in Michael Rutter’s PPA apparently have not been updated for the new release, and the old binaries are not ABI compatible with 3.5"....

So anyway, now travis is failing on some testthat stuff:
```
Failure: publishLandingPage works (@tests-publishLandingPage.R#18)  ──────
  any(grepl("climate-change-walleye-bass", index)) isn't true.
```

aappling-usgs:
Hey, cool, kudos for fixing the devtools issue!

PR #381, opened 2018-06-13
Fix for R 3.5

ldecicco-USGS:


ldecicco-USGS:
Fixes the landing page bug that is introduced in R 3.5

PR #386, opened 2018-07-19
customizing GA tracking snippet so custom dims start with values

wdwatkins:
I moved the functions for creating the `timestamp` and `sessionId` variables into the tracking snippet in the template.  The main idea here is to send a value for `sessionId` (`timestamp` is less necessary) along with the initial pageview hit.  Before (i.e. in the water use viz) we were relying on another event happening that would set the `sessionId` dimension.  But, if a user never interacted with the page, that never happened, and so a sessionId was never set.  With this change, the session ID should be set right from the start.

For some reason, Google doesn't give you NAs in custom dimensions, results just get left out that don't have them, which makes figuring out sessions that didn't do something a hassle.  

This works fine as far as GA is concerned, but is there any reason not to do this, load time performance or something?


wdwatkins:
Also, for `sessionId`, we could conceivably just reuse the variable created here and pass it in to all the functions that have GA events in them.  There isn't really a need to generate it more than once.

wdwatkins:
sounds good

aappling-usgs:
Oh, may not have been thinking right about `timestamp`...does that need to be generated on the fly for every new call to `gtag`? in that case, could you just define it as `'timestamp': getTimestamp()` within the arguments passed to `gtag` rather than assigning it to a variable first?

PR #387, opened 2018-07-20
namespace analytics vars, bump version

wdwatkins:


inst/templates/fullPage.mustache
```diff
@@ -19,18 +19,19 @@
       window.dataLayer = window.dataLayer || [];
       function gtag(){dataLayer.push(arguments)};
       gtag('js', new Date());
-      function getTimestamp() {return new Date().getTime().toString()}
-      function getSessionId() {return new Date().getTime() + '.' + Math.random().toString(36).substring(5)}
-      var sessionId = getSessionId();
-      var timestamp = getTimestamp();
+      var analytics = {
+        sessionId: (new Date().getTime() + '.' + Math.random().toString(36).substring(5)),
+        getTimestamp: function(){return new Date().getTime().toString()}
```
aappling-usgs: ooh, great idea!

wdwatkins: I believe you recommended it 😉

aappling-usgs: LOL - only partly =). I hadn't thought of putting the function itself into the `analytics` object

PR #389, opened 2018-10-02
kick off different webpack build configs for dev vs prod

lindsaycarr:
Tested this with `gage-conditions`. The dev bundle.js was 1.2 MB and the beta bundle.js was 250 KB.

```
vizlab::vizmake(webpack_build_cfg = "dev") # uses webpack --mode development
vizlab::vizmake(webpack_build_cfg = "beta") # uses webpack --mode production
```


mwernimont:
Nice!

PR #390, opened 2018-10-03
content.css addition

lindsaycarr:
Thanks to some quick screensharing with @mwernimont, we figured out why the header and footer spacing looked strange in `gage-conditions` (see https://github.com/USGS-VIZLAB/gage-conditions/issues/20). We updated `content.css` with the necessary style specs and increased the version.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: vizlab
 Type: Package
 Title: Utilities for building online data visualizations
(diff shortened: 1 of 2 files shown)


lindsaycarr:
Errors are still those same tests failing that have nothing to do with my tiny CSS change.

mwernimont:
I approve this PR

PR #393, opened 2018-10-29
drop imager from publish/thumbnail

wdwatkins:


wdwatkins:
Travis failed because of a Github API rate limit error 😕

